PR #7, opened 2019-11-26
No Internet Required + k8s v1.14.9

lamw:
* Changed to the latest k8s v1.14.9 via Kubeadm configuration file as the default version shipped w/PhotonOS 3.0 is still v1.14.6 which may not include all latest CVEs
* Pre-download all required Docker Containers, so internet connectivity is not required for initial VEBA setup
* Update ImagePullPolicy from Always to IfNotPresent which is required or k8s deployment spec will fail as it tries to pull Docker Container even though its already present
* Minor doc/typo fixes

### DIFF
--- a/files/setup.sh
+++ b/files/setup.sh
@@ -101,12 +101,12 @@ __CUSTOMIZE_PHOTON__
     # Setup k8s
     echo -e "\e[92mSetting up k8s ..." > /dev/console
     HOME=/root
-    kubeadm init --ignore-preflight-errors SystemVerification
+    kubeadm init --ignore-preflight-errors SystemVerification --skip-token-print --config /root/kubeconfig.yml
     mkdir -p $HOME/.kube
     cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
     chown $(id -u):$(id -g) $HOME/.kube/config
     echo -e "\e[92mDeloying kubeadm ..." > /dev/console
-    kubectl --kubeconfig /root/.kube/config apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl --kubeconfig /root/.kube/config version | base64 | tr -d '\n')"
+    kubectl --kubeconfig /root/.kube/config apply -f /root/weave.yaml
     kubectl --kubeconfig /root/.kube/config taint nodes --all node-role.kubernetes.io/master-
     echo -e "\e[92mStarting k8s ..." > /dev/console
     systemctl enable kubelet.service
--- a/photon-dev.json
+++ b/photon-dev.json
@@ -70,10 +70,24 @@
   "provisioners": [
     {
       "type": "shell",
+      "expect_disconnect" : true,
       "scripts": [
         "scripts/photon-settings.sh",
-        "scripts/photon-docker.sh",
-        "scripts/photon-openfaas.sh"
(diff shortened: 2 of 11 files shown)


scripts/photon-docker.sh
```diff
@@ -9,6 +9,7 @@
 echo '> Enabling Docker...'
 
 systemctl enable docker
+reboot
 
 echo '> Done'
```
embano1: Is this still called even though we add reboot?

files/setup.sh
```diff
@@ -101,12 +101,12 @@ __CUSTOMIZE_PHOTON__
...
     echo -e "\e[92mSetting up k8s ..." > /dev/console
     HOME=/root
-    kubeadm init --ignore-preflight-errors SystemVerification
+    kubeadm init --ignore-preflight-errors SystemVerification --skip-token-print --config /root/kubeconfig.yml
     mkdir -p $HOME/.kube
     cp -i /etc/kubernetes/admin.conf $HOME/.kube/config
     chown $(id -u):$(id -g) $HOME/.kube/config
     echo -e "\e[92mDeloying kubeadm ..." > /dev/console
-    kubectl --kubeconfig /root/.kube/config apply -f "https://cloud.weave.works/k8s/net?k8s-version=$(kubectl --kubeconfig /root/.kube/config version | base64 | tr -d '\n')"
+    kubectl --kubeconfig /root/.kube/config apply -f /root/weave.yaml
```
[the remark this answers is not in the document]
lamw: Looks like this was captured in the prior commit. See https://github.com/lamw/vcenter-event-broker-appliance/commit/7b411605ccab0fc006adc92e81df779e8e4bc3aa

http/photon-kickstart.json
```diff
@@ -3,7 +3,7 @@
     "password":
         {
             "crypted": false,
-            "text": "##FILL-IN-SECURE-PASSWORD##"
+            "text": "VMware1!"
```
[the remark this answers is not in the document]
lamw: Yes, this is mentioned in the docs. This changed as part of my testing :X and looks like that was captured. I could update this back to the system defaults

photon-dev.json
```diff
@@ -18,7 +18,7 @@
     "ovftool_deploy_vm_vcenter_server": "192.168.30.200",
     "ovftool_deploy_vm_vcenter_username": "administrator@vsphere.local",
     "ovftool_deploy_vm_vcenter_password": "VMware1!",
-    "ovftool_deploy_vm_network": "VM Network",
+    "ovftool_deploy_vm_network": "Isolated",
```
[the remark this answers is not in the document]
lamw: Same as before, it was captured as part of my testing and ultimately this is just an example. Users will need to update this with their own network

scripts/photon-docker.sh
lamw: Yes, this is part of the post-PhotonOS installation and is needed to be able to perform the docker pull

embano1: Are you also referring to this line?

> echo '> Done'

I.e. after a `reboot`, would that line be executed?

lamw: That's correct. This particular script will simply enable Docker, reboot and then come back and continue executing the remainder post-deploy scripts

PR #8, opened 2019-11-29
New example for host maintenance

doogleit:
This example disables alarms when a host is put in maintenance and re-enables alarms when it exits maintenance.
Signed-off-by: Doug Taliaferro <dougt42@gmail.com>


embano1:
@doogleit thx a ton for this super useful contribution and clearly written blog post! Excellent! 

Since we're capturing examples in a central [README](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples), can you please file another PR (since this is merged) and update the README? Don't worry about the missing Python counterpart.

PR #11, opened 2019-12-01
Update depreciated use of "faas" provider name

alexellis:
The provider name "faas" was depreciated 8 months ago and in
its place "openfaas" should be used. The vcenter-connector link
also needed to be updated in its description to match its name
upstream.

Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The vCenter Event Broker Appliance follows a highly modular approach, using Kube
 - Kubernetes ([Github](https://github.com/kubernetes/kubernetes))
 - Contour ([Github](https://github.com/projectcontour/contour))
 - OpenFaaS ([Website](https://www.openfaas.com/))
-- vCenter Connector ([Github](https://github.com/openfaas-incubator/vcenter-connector/))
+- OpenFaaS vcenter-connector ([Github](https://github.com/openfaas-incubator/vcenter-connector/))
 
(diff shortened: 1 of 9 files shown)


lamw:
Hi @alexellis - Happy to update the deprecated provider name, I wasn't aware this had changed. 

Regarding the description, can you clarify? We're simply following the name that's in the repo https://github.com/openfaas-incubator/vcenter-connector/ which clearly states `vcenter-connector`? Not sure this really warrants a change?

alexellis:
The author of the component has suggested updating the naming to match the upstream. That means it does warrant a change. The name is `openfaas-incubator/vcenter-connector` therefore you should refer to it as "OpenFaaS vcenter-connector" as suggested.

lamw:
@alexellis I'm trying to understand why this change is necessary when we've copied the text verbatim from the actual project readme and its linking directly to the project which IMHO is probably more important. 

I should also mention that the other referenced projects don't list out org/repo and simply use project name for simplicity purposes. Full details of each projects is already described in the [architecture.md](architecture.md) with description taken from each of the respective project.

alexellis:
I've updated the project README for your own clarity. I am struggling to understand why you have a problem with calling the OpenFaaS vcenter-connector "OpenFaaS vcenter-connector".

PR #18, opened 2019-12-11
Fixes to documentation and sample config files for openfaas

kremerpatrick:
The default provider name in stack.yml is 'faas', but the function fails to deploy with error "['openfaas'] is the only valid "provider.name" for the OpenFaaS CLI, but you gave: faas". Changing the default provider name to 'openfaas' fixes the problem.

The readme has a typo calling the command 'faas' when the command needs to be 'faas-cli'.

Changed the default provider gateway to a clearer placeholder - deploying VEBA out of the box requires HTTPS URL, and is not running on port 8080.


embano1:
Thank you Patrick! Great catches. Would you mind updating the other examples with the correct provider as well so we have this needed change in all examples?

embano1:
You can use the following git workflow for this:

```
# make changes to files
git add all
git commit -s --amend # create/update commit message with all changes
git push --force-with-lease
```

kremerpatrick:
OK I think I did it correctly.

PR #32, opened 2020-01-06
Add proxy settings to ova

### LINKED ISSUE #40: Allow network proxy settings during deployment

rguske:
Signed-off-by: Robert Guske <rguske@vmware.com>

Fixes #40 
  
Provides proxy settings to enable the provisioning of VEBA in environments that allows Internet connection through a proxy only. It is also necessary for functions that require an outgoing connection in such environments.
- [x] changes implemented
- [x] tested 
- validation that the provision of VEBA behind a proxy is successful

image
image

### DIFF
--- a/files/setup.sh
+++ b/files/setup.sh
@@ -31,6 +31,11 @@ else
     DNS_SERVER_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.dns")
     DNS_DOMAIN_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.domain")
     NTP_SERVER_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.ntp")
+    HTTP_PROXY_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.http_proxy")
+    HTTPS_PROXY_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.https_proxy")
+    PROXY_USERNAME_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.proxy_username")
+    PROXY_PASSWORD_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.proxy_password")
+    NO_PROXY_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.no_proxy")
     ROOT_PASSWORD_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.root_password")
     OPENFAAS_PASSWORD_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.openfaas_password")
     VCENTER_SERVER_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.vcenter_server")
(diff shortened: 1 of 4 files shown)


files/setup.sh
```diff
@@ -30,6 +30,10 @@ else
     GATEWAY_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.gateway")
     DNS_SERVER_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.dns")
     DNS_DOMAIN_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.domain")
+    PROXY_ENABLED_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.enableproxy")
```
lamw: This is actually not needed because if the user is explicitly filling out the proxy settings, they're telling us they want this configured. We just need to do some pre-checks and then fill in the right areas based on the input.

manual/photon.xml.template
```diff
@@ -30,6 +30,22 @@
             <Label>DNS Domain</Label>
             <Description>DNS Domain</Description>
         </Property>
+        <Property ovf:key="guestinfo.enableproxy" ovf:type="string" ovf:userConfigurable="true">
+            <Label>Enable Proxy</Label>
+            <Description>Type yes or no</Description>
+        </Property>
```
lamw: Not required as mentioned above

manual/photon.xml.template
```diff
@@ -30,6 +30,22 @@
...
+        <Property ovf:key="guestinfo.httpproxy" ovf:type="string" ovf:userConfigurable="true">
+            <Label>HTTP Proxy</Label>
+            <Description>Enter the address of your HTTP Proxy</Description>
+        </Property>
+        <Property ovf:key="guestinfo.httpsproxy" ovf:type="string" ovf:userConfigurable="true">
+            <Label>HTTPS Proxy</Label>
+            <Description>Enter the address of your HTTPS Proxy</Description>
+        </Property>
```
lamw: Its not clear what the input should be here ... is it just proxy server (e.g. proxy.vmware.com) or is it the full proxy string (e.g. http://user:pass@proxy.vmware.com:port)? 

One option is passthrough which is telling user they must specify the proxy format OR breakup each component (HTTP vs HTTPs, Port, User and Pass) and then re-construct the final URL which is then passed.

The latter is better customer experience and just means you'll have a few more OVF properties

* Proxy Server (HTTP)
* Proxy Server (HTTPS)
* Proxy Server Port
* Proxy Username
* Proxy Password
* No Proxy

manual/photon-dev.xml.template
```diff
@@ -30,6 +30,27 @@
...
+            <Label>Proxy Username (optional)</Label>
+            <Description>Username for the Proxy Server</Description>
+        </Property>
+        <Property ovf:key="guestinfo.proxy_password" ovf:password="true" ovf:type="string" ovf:userConfigurable="true" ovf:value="">
+            <Label>Proxy Password (optional)</Label>
+            <Description>Password for the Proxy User</Description>
+        </Property>
+            <Property ovf:key="guestinfo.no_proxy" ovf:type="string" ovf:userConfigurable="true" ovf:value="">
+            <Label>No Proxy</Label>
+            <Description>No Proxy for e.g. your internal domain suffix. Comma seperated (localhost, 127.0.0.1, domain.local)</Description>
```
lamw: `seperated` is miss-spelled, it should be `separated`

manual/photon.xml.template
```diff
@@ -30,9 +30,26 @@
...
+            <Label>Proxy Username (optional)</Label>
+            <Description>Username for the Proxy Server</Description>
+        </Property>
+        <Property ovf:key="guestinfo.proxy_password" ovf:password="true" ovf:type="string" ovf:userConfigurable="true" ovf:value="">
+            <Label>Proxy Password (optional)</Label>
+            <Description>Password for the Proxy User</Description>
+        </Property>
+        <Property ovf:key="guestinfo.no_proxy" ovf:type="string" ovf:userConfigurable="true">
+            <Label>No Proxy</Label>
+            <Description>No Proxy for e.g. your internal domain suffix. Comma seperated (localhost, 127.0.0.1, domain.local)</Description>
```
lamw: `seperated` is miss-spelled, it should be `separated`

rguske: Fixed it.
image

manual/photon-dev.xml.template
rguske: - [x] Fixed

rguske:
I´ve tested all changes locally with proxy settings set and not set.
_VEBA configuration at startup:_
image
_Configured Proxy settings:_
image

sofixa:
@lamw @rguske actually that won't suffice since Docker has its own proxy settings that need to be set as well: https://docs.docker.com/config/daemon/systemd/#httphttps-proxy

rguske:
Thanks for bringing it up @sofixa! I´ll look into it.

PR #43, opened 2020-01-10
AWS EventBridge sample

lamw:
Function leveraging PowerShell EventBridge cmdlet to forward VM Deleted Event from vCenter Server to AWS EventBridge

### DIFF
--- a/examples/powercli/eventbridge/README.md
+++ b/examples/powercli/eventbridge/README.md
@@ -1,0 +1,53 @@
+# AWS EventBridge Function
+
+## Description
+
+This function demonstrates using PowerShell and the EventBridge cmdlets to forward a vCenter Server Event to AWS EventBridge
+
+## Prerequisites
+
+* Already created custom [EventBridge Bus](https://docs.aws.amazon.com/eventbridge/latest/userguide/create-event-bus.html) and [EventBridge Rule](https://docs.aws.amazon.com/eventbridge/latest/userguide/create-rule-partner-events.html)
+* AWS Secret and Access Key with `AmazonEventBridgeFullAccess` policy
+
+## Instruction
+
+Step 1 - Initialize function, only required during the first deployment
+
+```
+faas-cli template pull
+```
+
+Step 2 - Update `stack.yml` and `eventbridge-secrets.json` with your environment information
+
+Step 3 - Build the function container
+
+```
+faas-cli build -f stack.yml
+```
+
+Step 4 - Push the function container to Docker Registry (default but can be changed to internal registry)
+
+```
+faas-cli push -f stack.yml
+```
+
+Step 5 - Deploy function to vCenter Event Broker Appliance
+
+```
+VEBA_GATEWAY=https://veba.primp-industries.com
+export OPENFAAS_URL=${VEBA_GATEWAY} # this is handy so you don't have to keep specifying OpenFaaS endpoint in command-line
+
+faas-cli login --username admin --password-stdin --tls-no-verify # login with your admin password
+faas-cli secret create eventbridge-secrets --from-file=eventbridge-secrets.json --tls-no-verify # create secret, only required once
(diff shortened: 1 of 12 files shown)


embano1:
Why does this PR have the Pod CIDR commit?

examples/powercli/eventbridge/handler/script.ps1
```diff
@@ -0,0 +1,32 @@
...
+$details = [pscustomobject] @{
+    CreatedTime = $json.CreatedTime;
+    UserName = $json.UserName;
+    VMName = $json.objectName;
+}
+
+$data = ($details | convertTo-Json).toString()
+
+$payload = New-Object Amazon.EventBridge.Model.PutEventsRequestEntry
+$payload.EventBusName = "VMware-VMC" #EventBride Bus Name
```
embano1: Why not make this an ENV so we don't hardcode/assume the bus name?

examples/powercli/eventbridge/template/powercli/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+WORKDIR /root
+
+USER root
+
+# Populate example here - i.e. "cat", "sha512sum" or "node index.js"
+SHELL [ "pwsh", "-command" ]
+ENV fprocess="xargs pwsh ./function/script.ps1"
+COPY function function
+# Set to true to see request in function logs
+ENV write_debug="true"
```
embano1: Redundant to `stack.yaml` ENV?

examples/powercli/eventbridge/eventbridge-secrets.json
```diff
@@ -0,0 +1,5 @@
+{
+    "AWS_REGION" : "us-west-2",
+    "AWS_ACCESS_KEY" : "XXXXXXXXXX",
+    "AWS_SECRET_KEY" : "XXXXXXXXXX"
+}
```
embano1: See comment below to make EVENT_BUS configurable

examples/powercli/eventbridge/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Instruction
+
+Step 1 - Initialize function, only required during the first deployment
+
+```
+faas-cli template pull
+```
+
+Step 2 - Update `stack.yml` and `eventbridge-secrets.json` with your environment information
```
embano1: Perhaps advice to change IMAGE as well?

files/setup.sh
```diff
@@ -37,6 +37,7 @@ else
     VCENTER_USERNAME_PROPERTY=$(vmtoolsd --cmd "info-get guestinfo.ovfEnv" | grep "guestinfo.vcenter_username")
```
embano1: All these files are from a different commit, please force push without this one.

examples/powercli/eventbridge/handler/script.ps1
lamw: Good point! Fixed and pushed the changes already

examples/powercli/eventbridge/template/powercli/Dockerfile
lamw: This was a copy from an earlier stack.yaml, but fixed and pushed the changes already

examples/powercli/eventbridge/eventbridge-secrets.json
lamw: Fixed and pushed the changes already

examples/powercli/eventbridge/README.md
lamw: I guess this depends if a user wants to just "consume" or "extend". My assumption was the former just to get started. Thoughts?

embano1: Well, next step says „build the container“ so it would fail on push using your Docker account. I think build is not needed here (ie, for just fn consumption image can remain but step 3 needs to be removed then)

PR #52, opened 2020-01-24
Updated PowerCLI tagging function to include a filter

kremerpatrick:
Added filter named TAG_FILTER in vcconfig.json

Only VMs that already have a tag attached that matches TAG_FILTER will be tagged with TAG_NAME


lamw:
Per discussion with @kremerpatrick, deferring filter for future as it has many more implications than just at the function level.

PR #58, opened 2020-03-03
Stricter linting on VMware Event Router

embano1:
- Use stricter linting during Docker build/release
- For local testing linting and clean Git dir is not enforced though


embano1:
> Build is failing due to the `golangci-lint` but I suppose this is to be fixed in another PR. Otherwise no comments this is interesting binary 😄

Exactly, PR incoming :)

PR #62, opened 2020-03-04
Fix Host MoRef processing for PCLI Host Maintenance Function

lamw:
@kremerpatrick found the initial v0.3 sample did not properly process the MoRef IDs for ESXi host, this PR fixes that sample. Patrick has also validated the function changes in his own environment to ensure things are working and the updated docker container has also been pushed. 

### DIFF
--- a/examples/powercli/hostmaint-alarms/handler/script.ps1
+++ b/examples/powercli/hostmaint-alarms/handler/script.ps1
@@ -12,7 +12,6 @@ if($env:function_debug -eq "true") {
 }
 
 $eventObjectName = $json.data.host.name
-$managedObjectReference = $json.data.host.type
 
 Set-PowerCLIConfiguration -InvalidCertificateAction Ignore  -DisplayDeprecationWarnings $false -ParticipateInCeip $false -Confirm:$false | Out-Null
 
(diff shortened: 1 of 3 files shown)


examples/powercli/hostmaint-alarms/handler/script.ps1
```diff
@@ -12,25 +12,30 @@ if($env:function_debug -eq "true") {
...
 # Get the vCenter AlarmManager
 $alarmManager = Get-View AlarmManager
-if ($json.topic -eq "entered.maintenance.mode") {
+if ($json.subject -eq "EnteredMaintenanceModeEvent") {
     # Disable alarm actions on the host
     Write-Host "Disabling alarm actions on host: $eventObjectName"
-    $alarmManager.EnableAlarmActions($managedObjectReference, $false)
+    $alarmManager.EnableAlarmActions($hostMoRef.MoRef, $false)
 }
 else {
```
embano1: Suggestion: in order to simplify this example how about just deploying on function (it's the same function anyways), subscribing to these two topics (Enter/Exit) and handling that in the logic?

Either way, to further harden the logic I'd add an `-eq` check for the `else` case as well checking for `ExitMaintenanceModeEvent`. If someone messes up the event subscription, this could cause alarms being enabled for any other event (as per the trigger annotation).

lamw: Good suggestion! Consolidated both

examples/powercli/hostmaint-alarms/stack.yml
```diff
@@ -14,16 +14,4 @@ functions:
...
-    environment:
-      write_debug: true
-      read_debug: true
-      function_debug: false
-    secrets:
-      - vc-hostmaint-config
-    annotations:
-      topic: ExitMaintenanceModeEvent
\ No newline at end of file
+      topic: EnteredMaintenanceModeEvent, ExitMaintenanceModeEvent
```
embano1: Can you please double-check whether this actually works, i.e. no `""` when multiple events are listed and space in between `", "`? `of-echo` should be fine to test with.

embano1: Also, I think the README needs to be updated as well.

PR #63, opened 2020-03-04
Add support for colorized log output

embano1:
Add support for colorized log output for better readability/troubleshooting.

image

### DIFF
--- a/vmware-event-router/internal/processor/aws_event_bridge.go
+++ b/vmware-event-router/internal/processor/aws_event_bridge.go
@@ -13,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/eventbridge"
 	"github.com/pkg/errors"
+	"github.com/vmware-samples/vcenter-event-broker-appliance/vmware-event-router/internal/color"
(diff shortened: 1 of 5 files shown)


vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -50,7 +51,7 @@ type eventPattern struct {
 // NewAWSEventBridgeProcessor returns an AWS EventBridge processor for the given
 // stream source.
 func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, source string, verbose bool, ms *metrics.Server) (Processor, error) {
-	logger := log.New(os.Stdout, "[AWS EventBridge] ", log.LstdFlags)
+	logger := log.New(os.Stdout, color.Purple("[AWS EventBridge] "), log.LstdFlags)
```
martindekov: Just a nit pick as I see OpenFaaS's color being the distinguishable blue. Why don't we go with AWS EventBridge's distinguishable orange ?

embano1: I thought I'd make all processors the same color, i.e. "purple". Right now we only support on processor. In the future, there might be multiple processors so having distinct colors would make sense. Will change metrics server from yellow to teal and use yellow for AWS.

Issue #65, opened 2020-03-04
Prioritize eth0 IP Address in vSphere UI


lamw:
If there are multiple network interfaces (which includes psuedo/logical interfaces like docker0, veth:*, etc) within a GuestOS, VMware Tools will randomly order the addresses which is not ideal if you want to quickly know the primary interface address. This behavior is expected but can be altered by specifying the priority as documented in this blog post [here](https://www.virtuallyghetto.com/2017/12/vmware-tools-10-2-0-enables-virtual-machine-vnic-exclusion-and-priority-re-ordering.html)

Below are two different VEBA deployments and you can see in the first deployment, eth0 address is being shown but in the second, the POD Network is actually displayed as the first. 
<img width="775" alt="Screen Shot 2020-03-04 at 10 03 28 AM" src="https://user-images.githubusercontent.com/602199/75908917-e6e48100-5dff-11ea-984a-a94d1023c16a.png">

![image](https://user-images.githubusercontent.com/602199/75908920-e77d1780-5dff-11ea-8795-17d97ca501e0.png)

The suggestion and fix is to prioritize eth0 by default in VEBA

PR #66, opened 2020-03-04
Ensuring eth0 interface is shown first in vSphere UI

lamw:
This is to resolve https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/65 

Validation:
- Deployed tools.conf configuration and ensure eth0 is listed first and then followed by either weave/docker0 for secondary interfaces. Hiding veth*, vxlan* and datapath interfaces by default as these aren't important to show in the vSphere UI

Screen Shot 2020-03-04 at 12 46 24 PM

### DIFF
--- a/scripts/photon-settings.sh
+++ b/scripts/photon-settings.sh
@@ -29,4 +29,14 @@ tdnf install -y \
 echo '> Creating directory for setup scripts'
 mkdir -p /root/setup
 
+echo '> Creating tools.conf to prioritize eth0 interface...'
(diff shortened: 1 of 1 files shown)


embano1:
Did you also deploy a fn to see whether the network stack/routing is still working as expected?

lamw:
@embano1 No, this change only affects what's show in our vSphere UI from VMware Tools. It does not affect anything else. Details is in the Issues which links to the blog post I wrote about this very topic :)

embano1:
LGTM then.

PR #67, opened 2020-03-05
Example gotag govmomi function

codegold79:


codegold79:
One big change that Daniel Linsley suggested was persisting the connection to Govmomi so I don’t log in/log out for each time gotag-fn is invoked. He said I can probably use something similar to the example at https://github.com/openfaas-incubator/golang-http-template Search for this section: Example persistent database connection pool between function calls 

Currently, I don't know how to go about persisting the connection. I had no idea OpenFaaS could store such data between invocations. I welcome suggestions.

examples/README.md
```diff
@@ -4,11 +4,11 @@ This page lists ready to use functions curated by the vCenter Event Broker commu
...
-| vSphere Tagging | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/python/tagging) | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/powercli/tagging) |
-| Send VM Configuration Changes to Slack | | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/powercli/hwchange-slack) |
-| Disable Alarms for Host Maintenance | | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/powercli/hostmaint-alarms) |
-| ESX Maximum transmission unit fixer | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/python/esx-mtu-fixer) | |
-| Datastore Usage Notification | | [Link](https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/master/examples/powercli/datastore-usage-email) |
+| Use Cases       | Python                                                                                                       | PowerCLI                                                                                                       | Golang-http |
```
embano1: "Golang-http" should be just "Golang"

examples/go/tagging/.gitignore
```diff
@@ -0,0 +1,3 @@
+template
+build
+vcconfig.toml
```
embano1: `vcconfig.toml` should be included in the repo as its required. For local development and to not expose sensitive data perhaps use a different file name and exclude that.

examples/go/tagging/README.MD
```diff
@@ -0,0 +1,130 @@
...
+urn:vmomi:InventoryServiceTag:019c0a9e-0672-48f5-ac2a-e394669e2916:GLOBAL
+```
+
+Take a note of the `urn:...` for `demotag1` as we will need it for the next steps.
+
+### Customize the function
+
+For security reasons, do not expose sensitive data. We will create a Kubernetes [secret](https://kubernetes.io/docs/concepts/configuration/secret/) which will hold the vCenter credentials and tag information. This secret will be mounted (by the appliance) into the function during runtime. The secret will need to be created via `faas-cli`.
+
+First, change the configuration file [vcconfig.toml](vcconfig.toml) holding your secret vCenter information located in this folder:
```
embano1: That's why we can't ignore `vcconfig.toml` via `.gitignore`

examples/go/tagging/README.MD
```diff
@@ -0,0 +1,130 @@
...
+```bash
+# set up faas-cli for first use
+export OPENFAAS_URL=https://VEBA_FQDN_OR_IP
+faas-cli login -p VEBA_OPENFAAS_PASSWORD --tls-no-verify # vCenter Event Broker Appliance is configured with authentication, pass in the password used during the vCenter Event Broker Appliance deployment process
+
+# now create the secret
+faas-cli secret create vcconfig --from-file=vcconfig.toml --tls-no-verify
+```
+
+**TIP:** If you need only to update the secret, the command is
```
embano1: Nice!

examples/go/tagging/README.MD
```diff
@@ -0,0 +1,130 @@
...
+
+```bash
+faas-cli secret update vcconfig --from-file=vcconfig.toml --tls-no-verify
+```
+
+**Note:** Delete the local `vcconfig.toml` after you're done with this exercise to not expose this sensitive information.
+
+Lastly, define the vCenter event which will trigger this function. Such function-specific settings are performed in the `gotag-fn.yml` file. Open and edit the `gotag-fn.yml` provided with in the examples/go/tagging directory. Change `gateway` and `topic` as per your environment/needs.
```
embano1: we generally use `stack.yml` across our examples for the stack configuration.

examples/go/tagging/README.MD
```diff
@@ -0,0 +1,130 @@
...
+
+**Note:** If you are running a vSphere DRS-enabled cluster the topic annotation above should be `DrsVmPoweredOnEvent`. Otherwise the function would never be triggered.
+
+### Deploy the function
+
+After you've performed the steps and modifications above, you can go ahead and deploy the function:
+
+```bash
+faas template store pull golang-http # only required during the first deployment
+faas-cli deploy -f gotag-fn.yml --tls-no-verify
```
embano1: See `stack.yml` comment above

examples/go/tagging/gotag-fn.yml
```diff
@@ -0,0 +1,16 @@
+version: 1.0
+provider:
+  name: openfaas
+  gateway: http://127.0.0.1:8080
```
embano1: Should be something like "https://veba.yourdomain.com"

examples/go/tagging/gotag-fn.yml
```diff
@@ -0,0 +1,16 @@
...
+functions:
+  gotag-fn:
+    lang: golang-http
+    handler: ./gotag-fn
+    image: fgold/gotag-fn:0.1.0
```
embano1: Needs to be replaced with vmware repo and image ID (please discuss with @lamw )

examples/go/tagging/gotag-fn.yml
```diff
@@ -0,0 +1,16 @@
...
+    environment:
+      write_debug: true
+      read_debug: true
+    secrets:
+      - gotag-vcconfig
```
embano1: This does not match the instructions in the README where the secret name is `vcconfig` (faas-cli secret create **vcconfig** --from-file=vcconfig.toml --tls-no-verify). Please use the name we also have in other examples here, i.e. vcconfig for consistency.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	log.Println("Logged out of Govmomi and Rest clients.")
+
+	return handler.Response{
+		Body:       []byte(message),
+		StatusCode: http.StatusOK,
+	}, nil
+}
+
+func loadTomlCfg() *vcConfig {
```
embano1: No `*T` return pointer needed.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	Value string `json:"Value"`
+	Type  string `json:"Type"`
+}
+
+// Handle a function invocation
+func Handle(req handler.Request) (handler.Response, error) {
+	ctx := context.Background()
+
+	// Retrieve vCenter and tag info from the config file.
+	vcc := loadTomlCfg()
```
embano1: vcc should be `cfg` or something similar.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Incoming request object.
+	var mor dataObj
```
embano1: how about `event` instead of `mor`?

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Config represents the toml vcconfig file
+type vcConfig struct {
+	Tag     tagInfo
+	VCenter vCenterAuth
+}
+
+// The request from vmware-event-router.
+type dataObj struct {
```
embano1: These types are really hard to read. You might want to copy the event payload structure from the vmware-event-router events package.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+// Handle a function invocation
+func Handle(req handler.Request) (handler.Response, error) {
+	ctx := context.Background()
+
+	// Retrieve vCenter and tag info from the config file.
+	vcc := loadTomlCfg()
+
+	// Incoming request object.
+	var mor dataObj
+	err := json.Unmarshal(req.Body, &mor)
```
embano1: Is `Body` an `io.Closer` that needs to be closed explicitly?

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+func loadTomlCfg() *vcConfig {
+	secretPath := "/var/openfaas/secrets/vcconfig"
+	secret, err := toml.LoadFile(secretPath)
+	if err != nil {
+		log.Println(err)
+	}
+
+	cfg := vcConfig{}
+	secret.Unmarshal(&cfg)
```
embano1: err not checked

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	// Fill information in the request into a govmomi type.
+	var moRef types.ManagedObjectReference
+	moRef.Value = mor.Data.Vm.Vm.Value
+	moRef.Type = mor.Data.Vm.Vm.Type
+
+	// Connect to vcenter with govmomi client
+	log.Println("Connecting to vCenter")
+	clients := connect(ctx, vcc.VCenter)
```
embano1: you're using a global var but closing the connection in each fn iteration. either make this a local variable then or persist the connection across invocation. also you don't check for errors when constructing the client. even though you log some errors to openfaas, it would be better to check for errors and return early.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	log.Println("Connection successful.")
+
+	// tag the VM moRef with tagID (URN)
+	clients.tagVM(ctx, moRef, vcc.Tag.Urn)
```
embano1: same here, no error handling

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Respond with a message that tagging of vm was complete
+	message := fmt.Sprintf(
+		"%v was tagged with %v",
+		moRef.Value,
+		vcc.Tag.Urn,
+	)
+	log.Println(string(message))
```
embano1: message is already of type string

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+	err = clients.Govmomi.Logout(ctx)
```
embano1: perhaps it's better to extract the client logic into its own package and use a defer on an exported method to logout from all sessions (or persist them across invocation and only logout when SIGTERM/SIGINTERRUPT)

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		log.Printf("Unable to log out of govmomi due to %v.", err)
+	}
+	err = clients.Rest.Logout(ctx)
```
embano1: since you don't return on errors before, if something goes wrong along the lines there will be a lot of errors.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		log.Printf("Unable to log out of rest due to %v.", err)
+	}
+	log.Println("Logged out of Govmomi and Rest clients.")
```
embano1: this log statement could be wrong if logout from vc succeeds, but not from the tagging API (not atomic)

examples/go/tagging/gotag-fn/handler.go
embano1: Also, no error handling, should return error

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+
+	cfg := vcConfig{}
+	secret.Unmarshal(&cfg)
+
+	return &cfg
+}
+
+// Clients needed to access vSphere APIs
+type Clients struct {
+	Govmomi *govmomi.Client
```
embano1: if you write a single logout() method you can keep the fields private.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	Rest    *rest.Client
+}
+
+// Connect saves connection information to clients.
+func connect(ctx context.Context, vca vCenterAuth) *Clients {
```
embano1: no error handling for the caller

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	u := url.URL{
+		Scheme: "https",
+		Host:   vca.Server,
+		Path:   "sdk",
+	}
+
+	log.Println("Get govmomi client.")
+	govc, err := govmomi.NewClient(ctx, &u, true)
+	if err != nil {
+		log.Println(err)
```
embano1: should return early here

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	c.Govmomi = govc
+
+	log.Println("Login to govmomi client")
+	u.User = url.UserPassword(vca.User, vca.Password)
+
+	err = c.Govmomi.Login(ctx, u.User)
```
embano1: I could be wrong, but if you pass url.URL with user/pass to newclient, doesnt that perform the login automatically?

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		log.Printf("ERROR: %v", err)
+	}
+
+	log.Println("Get rest client")
+	c.Rest = rest.NewClient(c.Govmomi.Client)
```
embano1: see comments on login/error handling above.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	log.Println("Login to rest client")
+	c.Rest.Login(ctx, u.User)
+
+	return &c
+}
+
+// Tag adds an existing tag to a VirtualMachine.
+func (clients *Clients) tagVM(ctx context.Context, vm types.ManagedObjectReference, tagID string) {
```
embano1: you mix methods and functions within the package for client handling. better make this separate package and use methods on the client to perform the various actions. you might be able to go with value semantics on either the client struct or fields.

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	// Get the tag manager which does the tagging.
+	m := tags.NewManager(clients.Rest)
+
+	// Attach tag to VM.
+	err := m.AttachTag(ctx, tagID, vm)
+	if err != nil {
```
embano1: this error should be returned to the caller and handled in the fn

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+		}, errors.New("managed Reference Object information must not be empty")
+	}
+
+	// Fill information in the request into a govmomi type.
+	var moRef types.ManagedObjectReference
+	moRef.Value = mor.Data.Vm.Vm.Value
+	moRef.Type = mor.Data.Vm.Vm.Type
+
+	// Connect to vcenter with govmomi client
+	log.Println("Connecting to vCenter")
```
embano1: log string should be lowercased

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+	handler "github.com/openfaas-incubator/go-function-sdk"
+	"github.com/pelletier/go-toml"
+	"github.com/vmware/govmomi"
+	"github.com/vmware/govmomi/vapi/rest"
+	"github.com/vmware/govmomi/vapi/tags"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// tagInfo is part of the toml vcconfig file
+type tagInfo struct {
```
embano1: could be simplified with

```go
type vcconfig struct {
	Tag struct {
		Action string
		URN    string
	}
	VCenter struct {
		Server   string
		Password string
		User     string
	}
}
```

Also see comments below on client handling - perhaps this struct could live in a separate client package and the logic be internalized

examples/go/tagging/gotag-fn/handler.go
```diff
@@ -0,0 +1,177 @@
...
+		log.Println(err)
+	}
+	c.Govmomi = govc
+
+	log.Println("Login to govmomi client")
+	u.User = url.UserPassword(vca.User, vca.Password)
+
+	err = c.Govmomi.Login(ctx, u.User)
+	if err != nil {
+		log.Printf("ERROR: %v", err)
```
embano1: inconsistent error printing. caps are usually not used in go. also consider wrapping and returning the error instead of handling them in these functions.

examples/go/tagging/gotag-fn/handler.go
embano1: > NewClient creates a new client from a URL. The client authenticates with the server with username/password before returning if the URL contains user information.

embano1:
> One big change that Daniel Linsley suggested was persisting the connection to Govmomi so I don’t log in/log out for each time gotag-fn is invoked. He said I can probably use something similar to the example at https://github.com/openfaas-incubator/golang-http-template Search for this section: Example persistent database connection pool between function calls
> 
> Currently, I don't know how to go about persisting the connection. I had no idea OpenFaaS could store such data between invocations. I welcome suggestions.

Yes, this makes sense and that's one reason why Alex created the middleware package for Go. I implemented this here: https://github.com/embano1/gotagfn/blob/8a7cdf7675d217294e7070ad2795a3cc2edeac18/gotag-fn/handler.go#L47

examples/go/tagging/gotag-fn/handler.go
codegold79: I agree--I really have to get up to date on how to name things. I'll take another look at your event router. I know it probably seems obvious to you that I should look at things like vmware-event-router, but in my mind, the connection wasn't made and I don't know we're naming the same things. LOL So, thanks for mentioning it :)

examples/go/tagging/gotag-fn/handler.go
codegold79: Neat--I didn't know we could do that 😸

examples/go/tagging/gotag-fn/handler.go
codegold79: I see that in the [golang-http middleware example](https://github.com/openfaas-incubator/golang-http-template), there is a `defer r.Body.Close()` in the example where `r` is the `*http.Request`. 

However, when I looked into why we need to do a Body.Close() at all, it is so a client can signal to the server that it is done with the response, and the connection can be used by the next client request. Close() doesn't close connections--it keeps them open, but makes it available for the next request.

What that means is as a server, we don't need to close requests. 

It's possible I totally misunderstood all the explanations I read. What do you think? Here are a couple of my sources:
(1) https://grokbase.com/t/gg/golang-nuts/138p15ng41/go-nuts-http-request-body-close
In the first source, Brad Fitzpatrick wrote, "You don't need to close it for server requests. It's unfortunate in some ways that we use Request for both client and server requests."

Caleb Spare said, "The only places where the net/http documentation mentions closing req.Body
is for the client usages (e.g., Get())."

(2) https://groups.google.com/forum/#!topic/golang-nuts/K9e0tyfU2hY
Here in the second source, Jesse McNelis writes, "HTTP is a request/response protocol. You can't send a response without a request and you can't receive a new request until you've processed the previous request.
The http pkg uses keep-alive to avoid having to make a new connection for each request, but this means that it needs to clear the previous request so it can wait for a new  one and that's what r.Body.Close() does."

examples/go/tagging/gotag-fn/handler.go
codegold79: fmt.Sprintf() returns a string...and I convert it to a string. Oops. 😆

examples/go/tagging/gotag-fn/handler.go
codegold79: Now that we're using Go 1.13, we can take advantage of the error package and the error.Wrap() function. I'll try to apply it everywhere I return an error--how exciting! 😸

examples/go/tagging/gotag-fn.yml
codegold79: I went ahead and replaced it with vmware repo. I'll talk to @lamw about getting the image uploaded once I made all the changes for this review and test out the code.

examples/go/tagging/gotag-fn/handler.go
codegold79: I made the changes as close to the vmware-event-router events package. What do you think?

examples/go/tagging/gotag-fn/handler.go
codegold79: I got rid of the log statement completely. Not sure if it is useful to anyone.

examples/go/tagging/.vcconfig
```diff
@@ -0,0 +1,8 @@
+[vcenter]
```
embano1: Why did you make that a hidden file instead of the `vcconfig.toml` we use across the examples?

examples/go/tagging/.vcconfig
```diff
@@ -0,0 +1,8 @@
...
+server = "10.0.0.1"
```
embano1: In your case, does your fn handle the "/sdk" endpoint correctly if not specified?

examples/go/tagging/README.MD
```diff
@@ -79,7 +79,7 @@ provider:
 functions:
   pytag-fn:
```
embano1: should be `gotag-fn`

examples/go/tagging/README.MD
```diff
@@ -67,7 +67,7 @@ faas-cli secret update vcconfig --from-file=vcconfig.toml --tls-no-verify
 
 **Note:** Delete the local `vcconfig.toml` after you're done with this exercise to not expose this sensitive information.
```
embano1: This is not in sync with the (hidden) config file you use above.

examples/go/tagging/handler/handler.go
```diff
@@ -17,58 +17,56 @@ import (
...
+		Password string
+		Server   string
+		User     string
+	}
 }
 
-// The request from vmware-event-router.
-type dataObj struct {
-	Data nest1 `json:"data"`
+// The structure for a Cloud Event.
```
embano1: These structures still throws me off when reviewing the code, especially since you use a different json tag for the "data" field (below). Since you already import the govmomi types, I think it makes sense to also leverage the `events.go` package (from the VMware Event Router folder) or copy the structure over to this file.

examples/go/tagging/handler/handler.go
```diff
@@ -17,58 +17,56 @@ import (
...
+	Vm vmEventArgument `json:"Vm"`
 }
-type nest2 struct {
-	Vm vmInfo `json:"Vm"`
+type vmEventArgument struct {
+	Vm managedObjectReference `json:"Vm"`
 }
-type vmInfo struct {
-	Value string `json:"Value"`
+type managedObjectReference struct {
```
embano1: Some line spaces missing here.

examples/go/tagging/handler/handler.go
```diff
@@ -17,58 +17,56 @@ import (
...
+		log.Error(err)
+		return
+	}
 
 	// Incoming request object.
-	var mor dataObj
-	err := json.Unmarshal(req.Body, &mor)
+	var event vmEvent
+	err := json.Unmarshal(req.Body, &event)
 	if err != nil {
```
embano1: you could log and then return.

examples/go/tagging/handler/handler.go
```diff
@@ -17,58 +17,56 @@ import (
...
 		return handler.Response{
 			Body:       nil,
 			StatusCode: http.StatusBadRequest,
 		}, err
 	}
 
-	if mor.Data.Vm.Vm.Value == "" || mor.Data.Vm.Vm.Type == "" {
+	if event.Event.Vm.Vm.Value == "" || event.Event.Vm.Vm.Type == "" {
```
embano1: Consider writing a unit test for your fn and check whether this comparison is correct. It could be that instead of the empty default value `""` of a type you get `nil` which this check won't hit.

examples/go/tagging/handler/handler.go
```diff
@@ -77,13 +75,13 @@ func Handle(req handler.Request) (handler.Response, error) {
 
 	// Fill information in the request into a govmomi type.
 	var moRef types.ManagedObjectReference
-	moRef.Value = mor.Data.Vm.Vm.Value
-	moRef.Type = mor.Data.Vm.Vm.Type
+	moRef.Value = event.Event.Vm.Vm.Value
```
embano1: See comment above on event structure, this is hard to read.

examples/go/tagging/handler/handler.go
```diff
@@ -17,58 +17,56 @@ import (
...
 // Handle a function invocation
 func Handle(req handler.Request) (handler.Response, error) {
 	ctx := context.Background()
 
 	// Retrieve vCenter and tag info from the config file.
-	vcc := loadTomlCfg()
+	cfg, err := loadTomlCfg()
+	if err != nil {
+		log.Error(err)
+		return
```
embano1: you don't send any http response here as you do in other cases.

examples/go/tagging/handler/handler.go
```diff
@@ -77,13 +75,13 @@ func Handle(req handler.Request) (handler.Response, error) {
...
 	var moRef types.ManagedObjectReference
-	moRef.Value = mor.Data.Vm.Vm.Value
-	moRef.Type = mor.Data.Vm.Vm.Type
+	moRef.Value = event.Event.Vm.Vm.Value
+	moRef.Type = event.Event.Vm.Vm.Type
 
 	// Connect to vcenter with govmomi client
-	log.Println("Connecting to vCenter")
+	log.Println("connect to vCenter")
 	clients := connect(ctx, vcc.VCenter)
```
embano1: Assuming some of the earlier comments on client handling are still not incorporated (e.g. error handling), so skipping this one

examples/go/tagging/handler/handler.go
```diff
@@ -77,13 +75,13 @@ func Handle(req handler.Request) (handler.Response, error) {
...
-	log.Println("Connection successful.")
+	log.Println("connection successful")
```
embano1: Really? :)

examples/go/tagging/handler/handler.go
```diff
@@ -94,35 +92,34 @@ func Handle(req handler.Request) (handler.Response, error) {
 		moRef.Value,
 		vcc.Tag.Urn,
 	)
-	log.Println(string(message))
+	log.Println(message)
 
 	err = clients.Govmomi.Logout(ctx)
 	if err != nil {
-		log.Printf("Unable to log out of govmomi due to %v.", err)
+		log.Printf("unable to log out of govmomi due to %v", err)
```
embano1: usually formatted error handling follows this pattern: `(what happend: %v, err)`

examples/go/tagging/handler/handler.go
```diff
@@ -141,24 +138,20 @@ func connect(ctx context.Context, vca vCenterAuth) *Clients {
...
 	}
 	c.Govmomi = govc
 
-	log.Println("Login to govmomi client")
 	u.User = url.UserPassword(vca.User, vca.Password)
 
 	err = c.Govmomi.Login(ctx, u.User)
 	if err != nil {
-		log.Printf("ERROR: %v", err)
+		log.Print(err)
```
embano1: not sure if `log.Println()` would be better to not mess up console output (newlines)

examples/go/tagging/handler/handler.go
```diff
@@ -141,24 +138,20 @@ func connect(ctx context.Context, vca vCenterAuth) *Clients {
...
 	govc, err := govmomi.NewClient(ctx, &u, true)
 	if err != nil {
 		log.Println(err)
 	}
 	c.Govmomi = govc
 
-	log.Println("Login to govmomi client")
 	u.User = url.UserPassword(vca.User, vca.Password)
 
 	err = c.Govmomi.Login(ctx, u.User)
```
embano1: Consider collapsing this into `NewClient` by adding user/pass to `url.URL` earlier.

examples/go/tagging/handler/handler.go
embano1: If you want a clean and simple approach (i.e. you only care about moref and nothing else in the CloudEvents payload):

```go
type incoming struct {
	Data types.Event `json:"data,omitempty"`
}
```

Then you need to check whether it's a proper event carrying vm information (note that not only VmEvent* can carry vm information, so the user must make sure to put in the correct event topics for this function, but this applies to all fn in general).

Pseudo code:

```go
var event incoming
	err = json.Unmarshal(b, &event)
	if err != nil {
		log.Println(err) // handle invalid JSON
                return
	}

if event.Data == nil || event.Data.Vm == nil {
	log.Println("is nil") // handle invalid VmEvent payload
	return
}
log.Printf("moref: %v", event.Data.Vm.Vm)
```

examples/go/tagging/.vcconfig
codegold79: Good point--the hidden file format will not work. I'll make it vcconfig.toml.

examples/go/tagging/.vcconfig
codegold79: Yup--so, I'll have the user give me the server address (e.g. "10.0.0.1"), and I'll append the "/sdk" path with the help of the net/url library.

examples/go/tagging/handler/handler.go
codegold79: Removed Govmomi.Login by adding User and Password info into the url.URL.

examples/go/tagging/handler/handler.go
codegold79: Ah, maybe connection was not successful. I'll get rid of that println.

examples/go/tagging/gotag-fn/handler.go
codegold79: I have created a global var to persist the connection. Also, error is being checked.

examples/go/tagging/gotag-fn/handler.go
codegold79: Client logic has been extracted.

examples/go/tagging/gotag-fn/handler.go
codegold79: I have implemented error wrapping.

examples/go/tagging/gotag-fn/handler.go
codegold79: I put the client in an external package. I'm thinking that means these fields should be public, but I'm not 100% sure.

examples/go/tagging/gotag-fn/handler.go
codegold79: Implemented global variable to persist connection and put client in its own package.

examples/go/tagging/handler/handler.go
codegold79: I get an error `./handler.go:157:46: cannot convert nil to type types.ManagedObjectReference` when I try to compare `event.Data.Vm.Vm` to `nil`. Maybe we can cover this after you see the unit tests.

examples/go/tagging/gotag-fn/handler.go
codegold79: Thanks very much for your suggestion. I am now using it:
```
type incoming struct {
	Data types.Event `json:"data,omitempty"`
}
```

examples/go/tagging/gotag-fn/handler.go
codegold79: Logout is only called with SIGTERM/SIGNINTERRUPT is sent.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
+	u := url.URL{
+		Scheme: "https",
+		Host:   cfg.VCenter.Server,
+		Path:   "sdk",
+	}
+	u.User = url.UserPassword(cfg.VCenter.User, cfg.VCenter.Password)
 
-	return cfg, nil
-}
+	// Currently, VEBA does not support signed certificates.
```
embano1: This is not correct. Your fn decides whether it will support self-signed certs. Consider making this an ENV var so users can opt in/out.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 
-	err = c.Govmomi.Login(ctx, u.User)
+func parseEventMoRef(req []byte) (types.ManagedObjectReference, error) {
+	var event incoming
+	var moRef types.ManagedObjectReference
+
+	err := json.Unmarshal(req, &event)
 	if err != nil {
-		log.Print(err)
+		return moRef, fmt.Errorf("parsing of request failed: %w", err)
```
embano1: should be `return nil, error...`

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 	}
 
-	c.Rest = rest.NewClient(c.Govmomi.Client)
-	c.Rest.Login(ctx, u.User)
+	if event.Data.Vm == nil || event.Data.Vm.Vm.Value == "" {
+		return moRef, errors.New("empty managed reference object")
```
embano1: same as above

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 	if err != nil {
+		log.Fatalf("retrieve managed reference object failed: %v", err)
 		return handler.Response{
 			Body:       nil,
 			StatusCode: http.StatusBadRequest,
 		}, err
 	}
 
-	if event.Event.Vm.Vm.Value == "" || event.Event.Vm.Vm.Type == "" {
+	log.Println("tag VM moRef with tagID (URN)")
```
embano1: Some (if not all) of these `log.Println` statements can be considered DEBUG information. You might want to only log them if an appropriate ENV var is set. Reason: if your fn gets popular and has a lot of invocations, it could flood the container host without providing any real value.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 		Body:       []byte(message),
 		StatusCode: http.StatusOK,
 	}, nil
 }
 
-func loadTomlCfg() (vcConfig, error) {
-	secretPath := "/var/openfaas/secrets/vcconfig"
-	secret, err := toml.LoadFile(secretPath)
+// vsConnect connects to vSphere govmomi API using information from vcconfig.toml.
+func vsConnect() error {
```
embano1: The func is called `vsConnect()` but strictly speaking it does more than that. Consider breaking cfg and client handling out (testability, separation of concerns).

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
+	}
+
+	return nil
 }
 
-// Connect saves connection information to clients.
-func connect(ctx context.Context, vca vCenterAuth) *Clients {
-	var c Clients
+func loadTomlCfg(path string) (vcConfig, error) {
+	cfg := vcConfig{}
```
embano1: You redeclare `cfg` here (either make local var or global - but not both) and names are mixed (global, local and return var), ie. not consistent.

examples/go/tagging/handler/handler.go
```diff
@@ -8,12 +8,13 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"os"
+	"os/signal"
+	"syscall"
 
+	"github.com/codegold79/vsphere/vsClient"
```
embano1: There's some issues in this pkg which could affect your fn code. E.g. Client construction could lead to half-initialized client, where the client is not correctly initialized but also != nil so your code would fail at a later stage. (hard to troubleshoot and test). Consider providing the client pkg as a separate pkg in this fn and not external dependency so it can be tested and reviewed together with this fn.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
-	// Connect to vcenter with govmomi client
-	log.Println("connect to vCenter")
-	clients := connect(ctx, vcc.VCenter)
-	log.Println("connection successful")
-
-	// tag the VM moRef with tagID (URN)
-	clients.tagVM(ctx, moRef, vcc.Tag.Urn)
-
 	// Respond with a message that tagging of vm was complete
 	message := fmt.Sprintf(
```
embano1: Prob easier to read when it's a oneliner.

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	for _, v := range tomls {
+		cfg, err := loadTomlCfg(v.configPath)
+		if err != nil {
+			t.Fatal(v.testDesc, failMark, err)
+		}
+		t.Log(v.testDesc, passMark)
+
+		if cfg.VCenter.Server == v.vcServer {
+			t.Logf("Server should be '%s'. %v", v.vcServer, passMark)
```
embano1: First time I run the tests I thought the test failed even though Go test was fine. The "should be" threw me off. Is this idiomatic Go test code? I have never seen this before (could be wrong here).

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,128 @@
...
+		testDesc   string
+		configPath string
+		vcServer   string
+		vcUser     string
+		vcPassword string
+		tagUrn     string
+		tagAction  string
+	}{
+		{
+			"Toml file should load correctly",
```
embano1: We don't have a test case for a malformed/missing/empty toml file :)

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,128 @@
...
+			"testdata/vcconfig2.toml",
+			"veba.local.corp",
+			"admin@vsphere.local",
+			"password1234",
+			"urn:vmomi:InventoryServiceTag:11f16f36-f5c4-4c29-b7d3-d9c7d12babe6:GLOBAL",
+			"detach",
+		},
+	}
+
+	for _, v := range tomls {
```
embano1: Convention for table-driven tests is usually []tests and tt as var

embano1: Also table-driven tests usually use structs for want/got to simplify the []tests structure (readability). This lets you compare structs instead of many strings and simplifies the test structure.

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,128 @@
...
+		} else {
+			t.Fatalf("Got tag action '%s', want: '%v'. %v", v.tagAction, cfg.Tag.Action, failMark)
+		}
+	}
+}
+
+// TestParseEventMoRef ensures that managed object reference value and type are
+// obtained by the event json that meets Cloud Event specifications.
+func TestParseEventMoRef(t *testing.T) {
+	var events = []struct {
```
embano1: Comments above apply here.

examples/go/tagging/gotag-fn/handler.go
embano1: I would have to dig into the specifics of how the server-side code in net/http handles closing the request body (this is a POST). I would assume, since you open the body for reading, it can still be considered good practice to close the body as soon as your logic is done, i.e. via `defer ...`. IMHO won't harm and signals that the authors thinks of preventing leakage of associated resources, e.g. file descriptor, etc.

examples/go/tagging/gotag-fn/handler.go
embano1: > Client logic has been extracted.

As per review comment below, consider making this part of this PR instead of a separate repo. There's some issues in the client code which could break this fn.

> Logout is only called with SIGTERM/SIGNINTERRUPT is sent.

Did you actually test that? Problem: if you use the same context and that is cancelled before calling `Logout` the govmomi call might fail with `context.Cancelled` and leak connections. Check the event router code for a workaround.

codegold79: I noticed `SIGTERM` in action last week, and I it does seem like, as you say, the context is cancelled before calling `Logout`. I thought at the time that maybe `Logout`, though not called in our situation, might be a backup path if the context failed to cancel. Is logout not necessary in our case? It doesn't seem like it if context is handling everything and session authentication is lost before we can properly log out. Here's what I saw when a pod shutdown (switching pods):

```
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 got signal: terminated,
2020/03/13 23:05:58 stderr: log out of vSphere
2020/03/13 23:05:58 SIGTERM received.. shutting down server in 10s
2020/03/13 23:05:58 Removing lock-file : /tmp/.lock
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 [entrypoint] SIGTERM received.. shutting down server in 10s
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 vSphere logout failed: govmomi api logout failed: ServerFaultCode: The session is not authenticated.
```

examples/go/tagging/gotag-fn/handler.go
codegold79: I don't think I open the body for reading. `Req` looks like it's just a struct, and I'm accessing the body, which is a slice of bytes. It's not the same on the client side (where the POST request is made), where the response body implements io.Writer.

Here's the function signature for handle(): 
```go
func Handle(req handler.Request) (handler.Response, error)
```

And here's the handler.Request type:
```go
type Request struct {
	Body        []byte
	Header      http.Header
	QueryString string
	Method      string
	Host        string
}
```
Source: https://github.com/openfaas-incubator/go-function-sdk/blob/70701da50a91/handler.go#L21

What does it mean to close a slice of bytes?

examples/go/tagging/gotag-fn/handler.go
codegold79: I'll definitely move the client code to this repo.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
+				StatusCode: http.StatusBadRequest,
+			}, fmt.Errorf("connect to vSphere failed: %w", err)
+		}
 	}
 
-	// Incoming request object.
-	var event vmEvent
-	err := json.Unmarshal(req.Body, &event)
+	// Set up os signal handling to log out of vSphere.
+	signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
```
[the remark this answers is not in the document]
codegold79: What is the purpose of signal handling here in `handler.go`? It seems that something else is handling os signals and shutting everything down, and we wouldn't need to do anything. Because you're right--I noticed that when a pod shut down, the authentication was also revoked and I didn't need to log out. 
```
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 got signal: terminated,
2020/03/13 23:05:58 stderr: log out of vSphere
2020/03/13 23:05:58 SIGTERM received.. shutting down server in 10s
2020/03/13 23:05:58 Removing lock-file : /tmp/.lock
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 [entrypoint] SIGTERM received.. shutting down server in 10s
2020/03/13 23:05:58 stderr: 2020/03/13 23:05:58 vSphere logout failed: govmomi api logout failed: ServerFaultCode: The session is not authenticated.
```

Should I just get rid of signal handling?

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
-	ctx := context.Background()
-
-	// Retrieve vCenter and tag info from the config file.
-	cfg, err := loadTomlCfg()
-	if err != nil {
-		log.Error(err)
-		return
+	// Connect to vSphere govmomi API if not already done. Also, load the required
+	// vcconfig.
+	if client == nil {
```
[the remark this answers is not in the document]
codegold79: Before you mentioned it, I've never used `sync.Once` and knew nothing about it. After reading about it just now, I think it's a great way to do what I'm trying to do. I'll implement it. Thanks for the suggestion.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 
-	// Incoming request object.
-	var event vmEvent
-	err := json.Unmarshal(req.Body, &event)
+	// Set up os signal handling to log out of vSphere.
+	signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
+	go handleSignal(ctx, sigCh)
+
+	// Retrieve the Managed Object Reference from the event.
+	moRef, err := parseEventMoRef(req.Body)
```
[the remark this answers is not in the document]
codegold79: Are we (handler.go) the server in this case? If so, then doesn't this part mean that we don't need to close anything?
> The Server will close the request body. The ServeHTTP Handler does not need to.

If we are client, then I'm very confused and my understanding of something is very wrong. But if we are client, yes, it does seem like we would be, as the docs say, "responsible for calling the Close method.for calling the Close method."

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
-	// Get the tag manager which does the tagging.
-	m := tags.NewManager(clients.Rest)
+func handleSignal(ctx context.Context, sigCh <-chan os.Signal) {
+	s := <-sigCh
+	log.Printf("got signal: %v,\nlog out of vSphere", s)
 
-	// Attach tag to VM.
-	err := m.AttachTag(ctx, tagID, vm)
+	err := client.Logout(ctx)
 	if err != nil {
```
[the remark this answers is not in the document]
codegold79: I'm thinking maybe getting rid of `Logout()` completely because the os signal doesn't transfer from `vmware-even-router` app to the OpenFaaS `handler.go`. And even if it did, all it would do is call `Logout()` which also looks unnecessary given that context shuts down `handler.go`.

What do you think?

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
 	}, nil
 }
 
-func loadTomlCfg() (vcConfig, error) {
-	secretPath := "/var/openfaas/secrets/vcconfig"
-	secret, err := toml.LoadFile(secretPath)
+// vsConnect connects to vSphere govmomi API using information from vcconfig.toml.
+func vsConnect() error {
+	var (
+		err        error
```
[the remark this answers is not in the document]
codegold79: Yes, I believe so. Declaring error ahead of time is needed so I can avoid short variable declaration, `:=` the first time err is assigned in 
```go
cfg, err = loadTomlCfg(configPath)
```

I have found that strange things happen when we try to use short variable declaration for a multi-variable return function and one of the variables to be assigned is a global variable. Here's where I show this unexpected behavior: https://play.golang.org/p/W1wmkpAAqTN

You would think that if we did `cfg, err := loadTomlCfg()`, that Go would realize `cfg` was a global variable and `err` was new. But it doesn't. It's as if it creates a new, shadow `cfg` variable. Because of this, I had to declare `err` ahead of time, and just use the `=` assignment operator.

examples/go/tagging/handler/handler.go
codegold79: Oops--I'll try to make `cfg` a local var. As for names, I'll try not to use `cfg` instead of `config`, regardless of the type of variable (global, local, returned).

examples/go/tagging/handler/handler_test.go
codegold79: I don't know if it's idiomatic or not. I'm still new enough that I still have to refer to textbooks to see how to set up my table tests. I like referring to Bill Kennedy's book Go In Action. Here's a preview of Chapter 9 where the "should" wording (and the checkmark, ballotX) is used: https://livebook.manning.com/book/go-in-action/chapter-9/10

But I don't think what Kennedy is doing is idiomatic Go since I've never seen this type of wording anywhere else. I wouldn't mind changing it to a different wording. I might switch to using wording like, "Server %s loaded as expected."

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
embano1: Perhaps the template itself is handling it? It's not an issue to have multiple subscribers to signals. Weird err response from the logout...this could be due to the context being cancelled and govmomi taking care of it? Not sure though...

You could try to replicate in a minimalistic environment where you cancel the context and see what happens?

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
embano1: We're "server" and likely net/http would help cleaning things up. BUT: it's considered good practice to always close io.ReadClosers after using them, e.g. via defer() instead of relying on someone else. Won't harm and trains muscle memory.

But again, the "body" we operate on is not of type io.ReadCloser anyways (I was confused by the variable names used)

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
embano1: OK, gotcha. My linter did not show this as shadowed. Either way, this can be solved with:

1) Using different variable for the cfg and set the global in a two-step approach
2) Set cfg to nil (or default value) because the shadowed variable would not affect the global anyways if it's an error returned (leaving the global in default/nil state - which is what you want anyways)

I'll let you decide.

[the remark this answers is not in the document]
embano1: (also just saw discussion below on global scope of this var ;)

examples/go/tagging/handler/handler_test.go
embano1: Ah, I guess Bill is using something like BDD here, which makes sense. I'm just not used to seeing this. Don't worry. (Btw: if you like nice tests, take a look at https://github.com/rakyll/gotest

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
embano1: Your fn sees the OS signal (different process, different signals/handlers). So it's good practice to keep (graceful) shutdown logic.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
-	cfg, err := loadTomlCfg()
-	if err != nil {
-		log.Error(err)
-		return
+	// Connect to vSphere govmomi API if not already done. Also, load the required
+	// vcconfig.
+	if client == nil {
+		err := vsConnect()
+		if err != nil {
+			log.Fatalf("connect to vSphere failed: %v", err)
```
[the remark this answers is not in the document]
codegold79: It looks like redundant text, but the verbs are different. For `log.Fatalf()` or `log.Printf()`, the verb is `%v`. But for wrapping the message into the error with `fmt.Errorf()`, the verb is `%w`. I'm not sure if it's worth creating another variable to save the message before the verb, though I can be convinced otherwise.

examples/go/tagging/handler/handler.go
```diff
@@ -31,140 +32,145 @@ type vcConfig struct {
...
-	Vm managedObjectReference `json:"Vm"`
-}
-type managedObjectReference struct {
-	Type  string `json:"Type"`
-	Value string `json:"Value"`
+type incoming struct {
+	Data types.Event `json:"data,omitempty"`
 }
 
+var (
```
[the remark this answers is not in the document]
codegold79: I made `client`, `cfgPath`, and `once` (sync.Once) global variables.
The others, `ctx`, `cfg`, and `sigCh` were made local variables.

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
codegold79: I ended up keeping `var err error` for a different reason: sync.Once does not allow for functions with returns. however, I still wanted to capture errors, so I declared the variable outside `once.Do()`.

examples/go/tagging/handler/handler.go
codegold79: I tried it, but `nil` is not the correct return type. The return expects a type.ManagedObjectReference type. Here's a simplified version on [Playground](https://play.golang.org/p/PTt3BZPR1VM). It produces this error: 
```
./prog.go:17:2: cannot use nil as type ManagedObjectReference in return argument
```

examples/go/tagging/handler/handler.go
codegold79: I think I have to return an empty struct, not `nil`, and they don't seem to be the same.

examples/go/tagging/handler/handler.go
codegold79: I'm not sure how to pass an environment variable to `handler.go`. Do you think it would be all right to add a section to vcconfig.toml, maybe something like the following?

```toml
[debug]
verbose = true
```

And then, I can do something like, 

```go
if verbose == true {
    log.Println("the error and success messages")
}
```

examples/go/tagging/handler/handler.go
codegold79: Per our discussion, I have moved client back to the `function` package, but in a different file than `handler.go`.

examples/go/tagging/gotag-fn/handler.go
codegold79: @embano1 spoke about this, and there is no need to close body.

examples/go/tagging/handler/handler.go
codegold79: Unit test for insufficient information for types.ManagedObjectReference was added. As far as I'm concerned, any event can trigger `gotag-fn` as long as it can send along VM value and VM type in the cloud event.

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
codegold79: Yeah, I can create a minimalist environment and see if I can replicate what is happening. 

Not sure if it helps, but I did find what you suggested to look for. I see that go-ttp is logging out in the Nimbus Events log: 
> 03/24/2020, 6:04:56 PM User VSPHERE.LOCAL\XXXXXXX@10.xxx.xx.xx logged out (login time: Wednesday, March 25, 2020 1:04:24 AM UTC, number of API invocations: 1, user agent: Go-http-client/1.1)

Will update my findings soon...

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
codegold79: Because `body` is not type io.ReadCloser, I'll not close it. It is confusing that req has a field called Body.

```go
 type Response struct {

	// Body the body will be written back
	Body []byte

	// StatusCode needs to be populated with value such as http.StatusOK
	StatusCode int

	// Header is optional and contains any additional headers the function response should set
	Header http.Header
}
```
Source: https://github.com/openfaas-incubator/go-function-sdk/blob/master/handler.go

examples/go/tagging/handler/handler.go
codegold79: I changed it such that `parseEventMoRef()` function returns a pointer to `type.MangedObjectReference`, just so I can use `nil` in case of error.

examples/go/tagging/handler/handler.go
codegold79: Same response as above.

examples/go/tagging/handler/handler.go
codegold79: vsConnect() now uses lock/unlock client and will connect if `client == nil`.

examples/go/tagging/gotag-fn/handler.go
codegold79: Client has been moved back into the `function` package and fields have been made private.

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
codegold79: In my code, every time `Handle()` function is called, I spin up a goroutine with `go signalHandle()`. Oops. 

The fix was to create another global variable, `once` of type `sync.Once`, and call `once.Do()` on `go signalHandle()`. That fixed the problem.

examples/go/tagging/handler/handler.go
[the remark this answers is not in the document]
codegold79: Signal handling has been fixed.

examples/go/tagging/handler/handler.go
```diff
@@ -191,11 +194,11 @@ func validateConfig(cfg vcConfig) error {
 func debug() bool {
 	verbose := os.Getenv("write_debug")
 
-	if verbose == "false" {
-		return false
+	if verbose == "true" {
```
embano1: nit: could be shortened to `if verbose {...}`
You might want to consider using a linter which highlights these.

examples/go/tagging/handler/handler.go
```diff
@@ -0,0 +1,244 @@
...
+	moRef.Type = event.Data.Vm.Vm.Type
+	moRef.Value = event.Data.Vm.Vm.Value
+
+	return &moRef, nil
+}
+
+func handleSignal(ctx context.Context) {
+	var sigCh = make(chan os.Signal)
+
+	signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
```
embano1: Just saw this is an unbuffered channel. It's recommended to make this a buffered channel (buffer of 2 in this case) -> https://golang.org/pkg/os/signal/#Notify

examples/go/tagging/handler/client.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// vsClient is a client for vSphere.
+type vsClient struct {
+	govmomi *govmomi.Client
+	rest    *rest.Client
+}
+
+func NewClient(ctx context.Context, u url.URL, insecure bool) (*vsClient, error) {
```
embano1: no need to export this function

examples/go/tagging/handler/handler.go
```diff
@@ -0,0 +1,244 @@
...
+
+	if verbose {
+		log.Printf("got signal: %v, log out of vSphere", s)
+	}
+
+	err := client.logout(ctx)
+	if verbose {
+		if err != nil {
+			log.Printf("vSphere logout failed: %v", err)
+		} else {
```
embano1: nit: instead of  `else` logic you can align the "happy path" (line of sight) with this pattern:

```go
	err := client.logout(ctx)
	if verbose {
		if err != nil {
			log.Printf("vSphere logout failed: %v", err)
			return
		}
		log.Println("logged out of govmomi and rest APIs")
	}
```

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,197 @@
...
+			"Event should be readable, even with minimal information",
+			"testdata/event2.json",
+			false,
+			vm{vmType: "VirtualMachine", Value: "vm-2"},
+		},
+		{
+			"Event should return error if VM type and value are null",
+			"testdata/eventErr1.json",
+			true,
+			vm{},
```
embano1: In case of error these should be nil not empty struct as per your function signature. The test still succeeds because you don't check for the moref result in that case.

examples/go/tagging/handler/handler_test.go
```diff
@@ -0,0 +1,197 @@
...
+	type vm struct {
+		vmType string
+		Value  string
+	}
+
+	var tests = []struct {
+		testDesc  string
+		jsonPath  string
+		expectErr bool
+		want      vm
```
embano1: should be `*vm` (see comment below)

examples/go/tagging/handler/handler.go
codegold79: My linter had alerted me to these lines, but I ignored it because I didn't think the logic would work. 

```
S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)
```

It suggested I do this (won't work):
```
return verbose
```

However, what you suggest will work.
```
if verbose {
   return true 
}

return false
```

Good eye @embano1 ! I'll go ahead and make that change.

examples/go/tagging/handler/handler.go
codegold79: I saw the linter suggest a buffer of 1, but I tested the code with no buffer and it worked. 

As you know, I am hesitant to add a buffer. However, I still don't fully understand how channels work. Because you have more experience and understanding, I'll go ahead and add that buffer of 2.

examples/go/tagging/handler/handler.go
codegold79: I keep forgetting returning nothing is allowed. Good idea!

examples/go/tagging/handler/handler.go
embano1: Ah, me complete #fail xD

examples/go/tagging/handler/handler.go
embano1: It is to reliably guarantee that signals can be processed. In your case you pick up the signal. But there can be times where signal processing is delayed (or you receive multiple of these as you specified) and that's where signals won't make it through. So the guidelines (also per stdlib docs) is to use a buffer (at least "1", depending on the number of signals to expect).

PR #71, opened 2020-03-07
Added unauthenticated SMTP and green status emails

kremerpatrick:
Allows SMTP_PASSWORD and SMTP_USERNAME to be blank in the JSON config - if both are blank, unauthenticated SMTP on port 25 will be sent. 

Added green alarm status for notification when alarm returns to normal.

Adjusted the email body language to match the alarm status.


kremerpatrick:
Screwed up the signing, have to do it again.

PR #82, opened 2020-03-17
vRO Function Example

lamw:
This example demonstrates using VEBA to trigger vRO Workflow using vRO REST API

### DIFF
--- a/examples/powershell/vro/handler/script.ps1
+++ b/examples/powershell/vro/handler/script.ps1
@@ -1,0 +1,91 @@
+
+# Process function Secrets passed in
+$SECRETS_FILE = "/var/openfaas/secrets/vro-secrets"
+$SECRETS_CONFIG = (Get-Content -Raw -Path $SECRETS_FILE | ConvertFrom-Json)
+
+# Process payload sent from vCenter Server Event
+$json = $args | ConvertFrom-Json
+if($env:function_debug -eq "true") {
+    Write-Host "DEBUG: json=`"$($json | Format-List | Out-String)`""
+}
+
+$vcenter = ($json.source -replace "https://","" -replace "/sdk","");
+$vmMoRef = $json.data.vm.vm.value;
+$vm = $json.data.vm.name;
+
+if($vmMoRef -eq "" -or $vm -eq "") {
+    Write-Host "Unable to retrieve VM Object from Event payload, please ensure Event contains VM result"
+    exit
+}
+
+# e.g. mgmt-vcsa-01.cpbu.corp/vm-2660
+$vroVmId = "$vcenter/$vmMoRef"
+
+# assumes following vRO Workflow https://github.com/kclinden/vro-vsphere-tagging as an example
+$body = @"
+{
+    "parameters":
+	[
+        {
+            "value": {
+                "sdk-object":{
+                    "type": "VC:VirtualMachine",
+                    "id": "$($vroVmId)"}
+                },
+            "type": "VC:VirtualMachine",
+            "name": "vm",
+            "scope": "local"
+        },
+        {
+            "value": {
+                "string":{
+                    "value": "$($SECRETS_CONFIG.TAG_CATEGORY_NAME)"
+                }
+            },
+            "type": "string",
+            "name": "categoryName",
+            "scope": "local"
+        },
+        {
+            "value": {
+                "string":{
+                    "value": "$($SECRETS_CONFIG.TAG_NAME)"
+                }
+            },
+            "type": "string",
+            "name": "tagToFind",
+            "scope": "local"
+        }
+	]
+}
+"@
+
+# Basic Auth for vRO execution
+$pair = "$($SECRETS_CONFIG.VRO_USERNAME):$($SECRETS_CONFIG.VRO_PASSWORD)"
+$bytes = [System.Text.Encoding]::ASCII.GetBytes($pair)
+$base64 = [System.Convert]::ToBase64String($bytes)
+$basicAuthValue = "Basic $base64"
+
+$headers = @{
+    "Authorization"="$basicAuthValue";
+    "Accept="="application/json";
+    "Content-Type"="application/json";
+}
+
+$vroUrl = "https://$($SECRETS_CONFIG.VRO_SERVER):443/vco/api/workflows/$($SECRETS_CONFIG.VRO_WORKFLOW_ID)/executions"
+
+if($env:function_debug -eq "true") {
+    Write-Host "DEBUG: vRoVmID=$vroVmId"
+    Write-Host "DEBUG: TagCategory=$($SECRETS_CONFIG.TAG_CATEGORY_NAME)"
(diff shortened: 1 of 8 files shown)


examples/powershell/vro/handler/script.ps1
```diff
@@ -0,0 +1,82 @@
...
+
+# Process payload sent from vCenter Server Event
+$json = $args | ConvertFrom-Json
+if($env:function_debug -eq "true") {
+    Write-Host "DEBUG: json=`"$($json | Format-List | Out-String)`""
+}
+
+$vcenter = ($json.source -replace "https://","" -replace "/sdk","");
+$vmMoRef = $json.data.vm.vm.value;
+$vm = $json.data.vm.name;
```
embano1: There's no safety checks here validating that these properties exist. Would it be better to throw an informative error (e.g. if someone triggers on the wrong event type) instead of blowing up the function (exception)? Or does PS handle missing (NULL) fields correctly and stop processing?

examples/powershell/vro/handler/script.ps1
```diff
@@ -0,0 +1,82 @@
...
+    Write-Host "DEBUG: vRoVmID=$vroVmId"
+    Write-Host "DEBUG: TagCategory=$($SECRETS_CONFIG.TAG_CATEGORY_NAME)"
+    Write-Host "DEBUG: TagName=$($SECRETS_CONFIG.TAG_NAME)"
+    Write-Host "DEBUG: vRoURL=`"$($vroUrl | Format-List | Out-String)`""
+    Write-Host "DEBUG: headers=`"$($headers | Format-List | Out-String)`""
+    Write-Host "DEBUG: body=$body"
+}
+
+Write-Host "Applying vSphere Tag: $($SECRETS_CONFIG.TAG_NAME) to VM: $vm ..."
+Invoke-Webrequest -Uri $vroUrl -Method POST -Headers $headers -SkipHeaderValidation -Body $body -SkipCertificateCheck
```
embano1: We should make these "insecure by default" flags (SkipCertificateCheck) configurable, e.g. using ENV vars.

examples/powershell/vro/stack.yml
```diff
@@ -0,0 +1,16 @@
...
+    handler: ./handler
+    image: vmware/veba-powershell-vro:latest
+    environment:
+      write_debug: true
+      read_debug: true
+      function_debug: false
+    secrets:
+      - vro-secrets
+    annotations:
+      topic: DrsVmPoweredOnEvent
```
embano1: Probably does not harm to set to `"DrsVmPoweredOnEvent,VmPoweredOnEvent"` so non-DRS users have a good experience.

examples/powershell/vro/template/powercli/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN echo "/usr/bin/pwsh" >> /etc/shells && \
+    echo "/bin/pwsh" >> /etc/shells && \
+    tdnf install -y powershell-6.2.3-1.ph3 unzip && \
+    pwsh -c "Set-PSRepository -Name PSGallery -InstallationPolicy Trusted" && \
+    find / -name "net45" | xargs rm -rf && \
+    tdnf erase -y unzip && \
+    tdnf clean all
+
+RUN mkdir -p /home/app
+USER root
```
embano1: IMHO not required to be set  (also below)

examples/powershell/vro/template/powercli/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN echo "Pulling watchdog binary from Github." \
+    && curl -sSL https://github.com/openfaas/faas/releases/download/0.9.14/fwatchdog > /usr/bin/fwatchdog \
+    && chmod +x /usr/bin/fwatchdog \
+    && cp /usr/bin/fwatchdog /root
+
+WORKDIR /root
```
embano1: Why the repetition?

examples/powershell/vro/template/powercli/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+USER root
```
embano1: Do we need to run as root? Some K8s environments block this which could cause bad UX.

examples/powershell/vro/vro-secrets.json
```diff
@@ -0,0 +1,8 @@
+{
+    "VRO_SERVER" : "vro.cpbu.corp",
+    "VRO_USERNAME": "william@cpbu.corp",
```
embano1: Probably better to use a generic name here, just so people don't think you leaked information :)

examples/powershell/vro/vro-secrets.json
```diff
@@ -0,0 +1,8 @@
...
+    "VRO_PASSWORD": "VMware1!",
```
embano1: I think it's better to use something like `REPLACE_ME`...

examples/powershell/vro/handler/script.ps1
lamw: 👍 Just did a check if either are NULL, then print and exit

examples/powershell/vro/handler/script.ps1
lamw: Done

examples/powershell/vro/stack.yml
lamw: Yup, I wonder if we need to be consistent with the other examples too

examples/powershell/vro/template/powercli/Dockerfile
lamw: Ah. This was merely copy/paste from Alex original templates :)

examples/powershell/vro/template/powercli/Dockerfile
lamw: Ah. This was merely copy/paste from Alex original templates :)

examples/powershell/vro/template/powercli/Dockerfile
lamw: Yea, not needed

examples/powershell/vro/vro-secrets.json
lamw: :)

examples/powershell/vro/handler/script.ps1
embano1: Final question: PS json module treats NULL and empty string "" as equivalent? I.e. if the string is not present it will be marshalled to `""` ?

examples/powershell/vro/stack.yml
embano1: Absolutely, will file an issue

lamw: 👍

examples/powershell/vro/handler/script.ps1
lamw: That's correct, rather than throw an error unlike other languages. Good/Bad :)

PR #90, opened 2020-03-23
Add Github action for stale issues/PRs

embano1:
After 60 days of inactivity, mark issues/PRs with label `stale` and
close after 5 days if no response is received (label not removed).


embano1:
Closing this as there seems to be an issue sometimes with Actions where it simply won't pick up a new workflow. Verified that this exact same file works in a personal repo and the other PR for the "greetings" workflow also works here. Perhaps need to revisit this in the future as it's going to remain a useful feature to clean up issues/PRs.

Issue #91, opened 2020-03-27
Introduce /etc/veba-version


lamw:
For informational + debugging/troubleshooting purposes, there's no easy way of obtaining the version of VEBA from within the Appliance. Today, the version is only available outside of the VM in the VM Annotation field as well as the filename. If either of those change, then the version is lost and we still do not have a way to identify the release from within. 

Proposal is to introduce `/etc/veba-version` file which would be created on the filesystem and simply contain the version such as `v0.4`. I'm curious if there's other information we may want to capture @embano1?

codegold79:
@lamw, would the event processor be placed in the `/etc/veba-version` file as well? If so, wouldn't the name be something less specific?

lamw:
After our chat, I think /etc/veba-release would be better name. It’s good point on whether this file should be name something else if it contains processor detail. Maybe Michael has some thoughts given the two pieces of info we should persist somewhere

embano1:
SGTM

PR #92, opened 2020-03-29
Add /etc/veba-release to include VEBA version, commit ID & event processor

lamw:
Closing https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/91

Version/Commit ID is generated from the build.sh script and passes that into Packer as environmental variables which is then used to construct the initial `/etc/veba-release` file. Upon first boot after a user configures VEBA, the processor type is appended to the file. 

Testing was done on both OpenFaaS and AWS Event Bridge and below are the outputs when cat'ing `/etc/veba-release`

```
# cat /etc/veba-release
Version: v0.3.0
Commit: 634263b
Processor: OpenFaaS
```

```
# cat /etc/veba-release
Version: v0.3.0
Commit: 634263b
Processor: EventBridge
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,10 +5,15 @@
 echo "Building OVA ..."
 rm -f output-vmware-iso/*.ova
 
+if [[ ! -z $(git status -s) ]]; then
+    echo "Dirty Git repository, please clean up any untracked files or commit them before building"
+    exit
+fi
+
 if [[ "$1" -gt "-1" ]] && [[ $1 == "dev" ]]; then
     echo "Applying packer build to photon-dev.json ..."
     packer build -var-file=photon-builder.json -var-file=photon-version.json photon-dev.json
 else
     echo "Applying packer build to photon.json ..."
-    packer build -var-file=photon-builder.json -var-file=photon-version.json photon.json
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -10,5 +10,5 @@ if [[ "$1" -gt "-1" ]] && [[ $1 == "dev" ]]; then
     packer build -var-file=photon-builder.json -var-file=photon-version.json photon-dev.json
 else
     echo "Applying packer build to photon.json ..."
-    packer build -var-file=photon-builder.json -var-file=photon-version.json photon.json
+    packer build -var "VEBA_VERSION=$(cat VERSION)" -var "VEBA_COMMIT=$(git rev-parse --short HEAD)" -var-file=photon-builder.json -var-file=photon-version.json photon.json
```
embano1: The only concern I have is that the embedded variable `VEBA_COMMIT` might resolve to a stale commit, e.g. dirty git repository where you make changes but those have not been committed. A typical workaround is to check for a clean git repo and otherwise fail the build to make sure all changes are committed (or stashed). This is enforced in the build of the event router for example.

Otherwise LGTM

Issue #75, opened 2020-03-11
Reorganize and clean up /root directory 


lamw:
All VEBA configuration files are created in /root which makes it a bit messy for anyone logging in for the first time. Will move all these files into /root/config

lamw:
I'm also going to remove the "dev" Packer build, it adds unnecessary complexity as additional files must be updated when new OVF properties are introduced. This was really added for my benefit for the initial development but with test automation scripts, its easy to deploy after a build.

PR #94, opened 2020-04-02
Restructure VEBA configuration files + Cleanup

lamw:
Closes https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/75 

**Changes:**

- Move all VEBA configuration files from /root to /root/config
- Removed "dev" Packer template and its dependent files which was optimized for testing purposes. This added an unnecessary complexity for others to contribute and with the sample deployment scripts in test directory, this is really not required anymore

**Validation:**

- Successfully deployed VEBA and confirm all configuration files have been moved to /root/config and ensure all containers are up and running.


embano1:
@lamw just realized that some of our docs still reference to the old location. we need to sync docs with this change, e.g.  https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/development/docs/8-veba-troubleshooting.md

Issue #69, opened 2020-03-06
Honor AWS EventBridge Put Limits


embano1:
Requires batching (max 10 entries per `PutRequest`) and rate limiting (per region/account) as described in the limits: https://docs.aws.amazon.com/eventbridge/latest/userguide/cloudwatch-limits-eventbridge.html

martindekov:
Hey @embano1 for context I have couple of questions as per the current implementation:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/4d3027881bf9c0d5de2847385006b19373fe003e/vmware-event-router/internal/processor/aws_event_bridge.go#L198

So first questions is, does events arrive in batch from vCenter. I see this line of code counts them:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/4d3027881bf9c0d5de2847385006b19373fe003e/vmware-event-router/internal/stream/vcenter.go#L129

I though we can receive only one event form vCenter per request. I think this proves me wrong? I suppose couple of things in the vCenter can happen in parallel? So batch of events are send to the router. If this is the case we don't explicitly set max 10 entries here:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/4d3027881bf9c0d5de2847385006b19373fe003e/vmware-event-router/internal/processor/aws_event_bridge.go#L208

So essentially we set the full stack to event bridge which we receive from vCenter is that so?

Do you suggest to spawn a goroutine which will do this:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/4d3027881bf9c0d5de2847385006b19373fe003e/vmware-event-router/internal/processor/aws_event_bridge.go#L208

and will make sure we will send the events to event bridge at max 10 events. So:

1. 15 events from vCenter arrive
2. first 10 events are sent to event bridge
3. last 5 events are kept
4. 3 events arrive from vCenter
5. the previous 5 are appended with the current 3 so 8 events are sent to event bridge

As currently as far as I understood it we send everything we received from vCenter. Sorry for the long post I want to be explicit with the approach especially in the beginning of my contribution :smile:

embano1:
Thank you Martin for looking into this and your contribution! All good questions, let me answer with my thoughts on how I thought about addressing this `TODO` item.

I think this issue is actually two issues (at least related ones):

1. Honor batch limits per `PutEvents` call
2. Honor region/account quotas (i.e. rate limits) per `PutEvents` call

[...]

PR #95, opened 2020-04-02
Add input entry batching

martindekov:
Adding input entry batching which means that
we will no longer send all events with single
PutEvents call, but rather will batch the
events and will send at most 10 with single
PutEvents call, thus honoring the AWS limits
of at most 10 entries per request. Also adding
execution of tests as step of the docker build

Closes: #69 

Signed-off-by: Martin Dekov <mvdekov@gmail.com>

## Testing done:

1. Built image
2. Added unit tests

>Note: haven't tested in live environment but can help someone deploy the change if someone has live AWS environment

### DIFF
--- a/vmware-event-router/Makefile
+++ b/vmware-event-router/Makefile
@@ -46,7 +46,7 @@ gofmt:
 	@test -z "$(shell gofmt -s -l -d -e ./cmd | tee /dev/stderr)"
 
 test: gofmt
-	GORACE=history_size=5 go test -race -timeout $(TIMEOUT)s $(TESTPKGS)
+	GORACE=history_size=5 go test -race -timeout -cover $(TIMEOUT)s $(TESTPKGS) 
 
 tag: 
 	$(info Make: Tagging image "$(IMAGE_NAME)" with "$(BUILD_TAG)" and "$(LATEST_TAG)".)
--- a/vmware-event-router/internal/processor/aws_event_bridge.go
+++ b/vmware-event-router/internal/processor/aws_event_bridge.go
@@ -28,6 +28,7 @@ const (
 	authMethodAWS  = "access_key"    // only this method is supported by the processor
 	resyncInterval = time.Minute * 5 // resync rule patterns after interval
 	pageLimit      = 50              // max 50 results per page for list operations
+	batchSize      = 10              // max 10 input events per batch sent to AWS
 )
 
 // awsEventBridgeProcessor implements the Processor interface
@@ -115,7 +116,7 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 			return nil, errors.Wrap(err, "could not list event bridge rules")
 		}
 
-		arnLoop:
+	arnLoop:
 		for _, rule := range rules.Rules {
 			switch {
 			case *rule.Arn == ruleARN:
@@ -173,36 +174,40 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 // throttling/batching
 // https://docs.aws.amazon.com/eventbridge/latest/userguide/cloudwatch-limits-eventbridge.html#putevents-limits
 func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObjectReference, baseEvent []types.BaseEvent) error {
-	input, err := awsEventBridge.createPutEventsInput(baseEvent)
+	batchInput, err := awsEventBridge.createPutEventsInput(baseEvent)
 	if err != nil {
 		awsEventBridge.Printf("could not create PutEventsInput for event(s): %v", err)
 		return nil
 	}
 
 	// nothing to send
-	if len(input.Entries) == 0 {
+	if len(batchInput) == 0 {
 		return nil
 	}
 
-	// TODO: investigate limits on number/size of entries in a single put
-	resp, err := awsEventBridge.PutEvents(&input)
-	if err != nil {
-		awsEventBridge.Printf("could not send event(s): %v", err)
-		return nil
-	}
-	if awsEventBridge.verbose {
-		awsEventBridge.Printf("successfully sent event(s) from source %s: %+v", awsEventBridge.source, resp)
+	for idx, input := range batchInput {
+		resp, err := awsEventBridge.PutEvents(&input)
+		if err != nil {
+			awsEventBridge.Printf("could not send event(s) for batch %d: %v", idx, err)
+			continue
+		}
+		if awsEventBridge.verbose {
+			awsEventBridge.Printf("successfully sent event(s) from source %s: %+v batch: %d",
+				awsEventBridge.source,
+				resp,
+				idx)
+		}
 	}
 	return nil
 }
 
-func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []types.BaseEvent) (eventbridge.PutEventsInput, error) {
-	// TODO: Array Members: Minimum number of 1 item. Maximum number of 10 items. for []*eventbridge.PutEventsRequestEntry{}
-	// https://github.com/pacedotdev/batch
+func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []types.BaseEvent) ([]eventbridge.PutEventsInput, error) {
 	awsEventBridge.mu.Lock()
 	defer awsEventBridge.mu.Unlock()
 
-	input := eventbridge.PutEventsInput{
+	batch := []eventbridge.PutEventsInput{}
+
+	tmpInput := eventbridge.PutEventsInput{
 		Entries: []*eventbridge.PutEventsRequestEntry{},
 	}
 
@@ -221,9 +226,8 @@ func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []
 		cloudEvent := events.NewCloudEvent(event, eventInfo, awsEventBridge.source)
 		jsonBytes, err := json.Marshal(cloudEvent)
 		if err != nil {
-			return eventbridge.PutEventsInput{}, errors.Wrapf(err, "could not marshal cloud event for vSphere event %d from source %s", event.GetEvent().Key, awsEventBridge.source)
+			return nil, errors.Wrapf(err, "could not marshal cloud event for vSphere event %d from source %s", event.GetEvent().Key, awsEventBridge.source)
 		}
-
 		jsonString := string(jsonBytes)
 		entry := eventbridge.PutEventsRequestEntry{
 			Detail:       aws.String(jsonString),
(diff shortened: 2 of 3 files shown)


vmware-event-router/Dockerfile
```diff
@@ -18,6 +18,8 @@ COPY internal internal
 RUN ./bin/golangci-lint run ./...
 RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix nocgo -ldflags="-X main.version=${VERSION} -X main.commit=${COMMIT}" -o vmware-event-router cmd/main.go
 
+RUN go test $(go list ./...) -cover
```
embano1, line 21: can we move "cover" into the `Makefile` and `test` target please?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -115,7 +115,7 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 			return nil, errors.Wrap(err, "could not list event bridge rules")
 		}
 
-		arnLoop:
+	arnLoop:
```
embano1, line 118: Assuming this was a formatting issue on my end, correct?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -173,36 +173,38 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-	if err != nil {
-		awsEventBridge.Printf("could not send event(s): %v", err)
-		return nil
-	}
-	if awsEventBridge.verbose {
-		awsEventBridge.Printf("successfully sent event(s) from source %s: %+v", awsEventBridge.source, resp)
+	for _, input := range batchInput {
+		resp, err := awsEventBridge.PutEvents(&input)
+		if err != nil {
+			awsEventBridge.Printf("could not send event(s): %v", err)
```
embano1, line 190: Are we deciding on the first failure stops the whole iteration (batches)?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -173,36 +173,38 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
+			return nil
+		}
+		if awsEventBridge.verbose {
+			awsEventBridge.Printf("successfully sent event(s) from source %s: %+v", awsEventBridge.source, resp)
```
embano1, line 194: Should we indicate a batch number or something else for better troubleshooting?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -173,36 +173,38 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-	// https://github.com/pacedotdev/batch
+func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []types.BaseEvent) ([]eventbridge.PutEventsInput, error) {
 	awsEventBridge.mu.Lock()
 	defer awsEventBridge.mu.Unlock()
 
-	input := eventbridge.PutEventsInput{
+	batch := []eventbridge.PutEventsInput{}
+	batchSize := 10
+
+	tmpInput := eventbridge.PutEventsInput{
```
embano1, line 207: I think this can be scoped to the loop so you also save the reset on line 242?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -173,36 +173,38 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []types.BaseEvent) (eventbridge.PutEventsInput, error) {
-	// TODO: Array Members: Minimum number of 1 item. Maximum number of 10 items. for []*eventbridge.PutEventsRequestEntry{}
-	// https://github.com/pacedotdev/batch
+func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []types.BaseEvent) ([]eventbridge.PutEventsInput, error) {
 	awsEventBridge.mu.Lock()
 	defer awsEventBridge.mu.Unlock()
 
-	input := eventbridge.PutEventsInput{
+	batch := []eventbridge.PutEventsInput{}
+	batchSize := 10
```
embano1, line 205: can we make this a constant in this package for better documentation please?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -221,23 +223,32 @@ func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []
 		cloudEvent := events.NewCloudEvent(event, eventInfo, awsEventBridge.source)
 		jsonBytes, err := json.Marshal(cloudEvent)
 		if err != nil {
-			return eventbridge.PutEventsInput{}, errors.Wrapf(err, "could not marshal cloud event for vSphere event %d from source %s", event.GetEvent().Key, awsEventBridge.source)
+			return []eventbridge.PutEventsInput{}, errors.Wrapf(err, "could not marshal cloud event for vSphere event %d from source %s", event.GetEvent().Key, awsEventBridge.source)
```
embano1, line 226: Would it be ok to return `nil` instead of the slice?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -221,23 +223,32 @@ func (awsEventBridge *awsEventBridgeProcessor) createPutEventsInput(baseEvent []
...
 			Source:       aws.String(cloudEvent.Source),
 			DetailType:   aws.String(cloudEvent.Subject),
 		}
-		input.Entries = append(input.Entries, &entry)
+		tmpInput.Entries = append(tmpInput.Entries, &entry)
 
 		// update metrics
 		awsEventBridge.stats.Invocations[eventInfo.Name]++
+
+		if idx%batchSize == 0 && idx != 0 {
```
embano1, line 240: Please help me to understand the `&& idx != 0`

vmware-event-router/internal/processor/aws_event_bridge_test.go
```diff
@@ -0,0 +1,93 @@
...
+		desiredEntries int
+	}{
+		{
+			title:          "13 events 2 batches",
+			baseEvents:     createBaseEventsMock(13),
+			desiredBatches: 2,
+			desiredEntries: 13,
+		},
+		{
+			title:          "10 events 1 barch",
```
embano1, line 24: typo

vmware-event-router/internal/processor/aws_event_bridge_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"testing"
+
+	"github.com/vmware-samples/vcenter-event-broker-appliance/vmware-event-router/internal/metrics"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+func Test_batching_createPutEventsInput(t *testing.T) {
+	tests := []struct {
```
embano1, line 11: can we have a test case with events in the `[]BaseEvent` slice that we don't have a subscription for? E.g. to test a large number of events in the incoming event slice but only a subsection of them will make it into the final batch(es) sent to AWS.

vmware-event-router/internal/processor/aws_event_bridge_test.go
martindekov: Nice catch :smile: changed :heavy_check_mark:

vmware-event-router/internal/processor/aws_event_bridge_test.go
martindekov: Done :+1:

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Yeah so this logic here is used only if the batches are more than 10. In case for example the input entries are 0-9 we want a single batch which is created on line 247. This is just that 0 % 0 == 0 and we will have 2 batches instead of the desired one. So if we have 3 input entries and remove this we will have 1 batch with 1 input entry and then a second batch with 3 input entries when we only want 1 batch with 3 input entries.

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Done :+1:

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Yeah Michael, actually this is used when the loop ends to create batch when the input entries are less than 10.

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Done :+1:

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Done :heavy_check_mark: good point there

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: I miscalculated that you are right we should not return, instead I have changed with `continue` :+1:

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: I have formatter on save, probably this is the right formatting though I am not sure.

vmware-event-router/Dockerfile
martindekov: Done :heavy_check_mark:

vmware-event-router/internal/processor/aws_event_bridge_test.go
embano1: Sorry, I probably miscommunicated. What I was asking for is different types of events in the same `[]BaseEvent`. E.g. having `VmPoweredOnEvent` and `VmPoweringOnWithCustomizedDVPortEvent` mixed but a batch entry should only be created for one of these (e.g. pattern map only wants `VmPoweredOnEvent`). This is very possible and should be tested as well.

martindekov: Yeah Michael, this is a totally a valid test case. I added a method which mixes events of two types and sends them to the processor, verifying that only the desired event (to which the processor is subscribed) is put into the batches which are to be sent to the AWS Event Bridge :+1:

vmware-event-router/Makefile
```diff
@@ -46,7 +46,7 @@ gofmt:
 	@test -z "$(shell gofmt -s -l -d -e ./cmd | tee /dev/stderr)"
 
 test: gofmt
-	GORACE=history_size=5 go test -race -timeout $(TIMEOUT)s $(TESTPKGS)
+	GORACE=history_size=5 go test -race -timeout $(TIMEOUT)s $(TESTPKGS) -cover
```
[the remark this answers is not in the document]
martindekov: Yeah, good idea, Done :heavy_check_mark:

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -173,36 +174,40 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-	if err != nil {
-		awsEventBridge.Printf("could not send event(s): %v", err)
-		return nil
-	}
-	if awsEventBridge.verbose {
-		awsEventBridge.Printf("successfully sent event(s) from source %s: %+v", awsEventBridge.source, resp)
+	for idx, input := range batchInput {
+		resp, err := awsEventBridge.PutEvents(&input)
+		if err != nil {
+			awsEventBridge.Printf("could not send event(s): %v for batch: %d", err, idx)
```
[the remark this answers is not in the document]
martindekov: Places swapped, indeed looking for the batch at the end of the error might be confusing :+1:

martindekov:
Thanks for the feedback Michael, added changes as you proposed :+1:

martindekov:
Yep, will squash no problems, also will provide with the image containing the change as I haven't tested in in a live environmen (AWS Event Bridge setup).

martindekov:
Hey @embano1 squashed the commits and pushed an image in Dockerhub with the change: `martindekov/vmware-event-router:0.4.0-rc1`

PR #101, opened 2020-04-04
Add Otto to start page

embano1:


embano1:
Let me know if it should be centred.

Issue #102, opened 2020-04-06
Issues deploying to air-gapped environment.


ryan-jan:
As discussed in the Slack channel, I am having issues deploying v0.3 to an air-gapped network as some images are not available at the correct version locally. For example `weave-kube` and `weave-npc` v2.6.1 are not available.

```
root@rktest01 [ ~ ]# docker image list
REPOSITORY                           TAG                 IMAGE ID            CREATED             SIZE
vmware/veba-event-router             latest              40af0186f445        3 weeks ago         119MB
k8s.gcr.io/kube-proxy                v1.14.9             636041c2a488        4 months ago        82.1MB
k8s.gcr.io/kube-apiserver            v1.14.9             5811259ed0c9        4 months ago        209MB
k8s.gcr.io/kube-controller-manager   v1.14.9             07193a77f264        4 months ago        157MB
k8s.gcr.io/kube-scheduler            v1.14.9             0f036524b7a2        4 months ago        81.6MB
weaveworks/weave-npc                 2.6.0               5105e13e253e        5 months ago        34.9MB
weaveworks/weave-kube                2.6.0               174e0e8ef23d        5 months ago        114MB

root@rktest01 [ ~ ]# kubectl get events -n kube-system
38m         Normal    LeaderElection      endpoints/kube-scheduler               rktest01_1796a1b4-77e8-11ea-8ed2-005056b993a7 became leader
38m         Normal    Scheduled           pod/weave-net-5tqhs                    Successfully assigned kube-system/weave-net-5tqhs to rktest01
37m         Normal    Pulling             pod/weave-net-5tqhs                    Pulling image "docker.io/weaveworks/weave-kube:2.6.1"
36m         Warning   Failed              pod/weave-net-5tqhs                    Failed to pull image "docker.io/weaveworks/weave-kube:2.6.1": rpc error: code = Unknown desc = Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
36m         Warning   Failed              pod/weave-net-5tqhs                    Error: ErrImagePull
36m         Normal    Pulling             pod/weave-net-5tqhs                    Pulling image "docker.io/weaveworks/weave-npc:2.6.1"
[...]

PR #106, opened 2020-04-09
Fix VMware Event Router image pull to support air-gap scenario

lamw:
> Summary

Updated the logic on how the VMware Event Router image is pulled per https://github.com/vmware-samples/vcenter-event-broker-appliance/pull/105 and fixed Event Router K8s YAML to reflect the image version + imagePullPolicy for airgap/non-internet requirement

> Resolves

Close: https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/102

> Testing

**Internal:** Deployed VEBA w/OpenFaa Processor against vSphere 7.0 build and verified deployment was successful and all pods were running as expected

**External:** Provided a build w/fixes to customer which has been verified https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/102#issuecomment-611537305

### DIFF
--- a/files/setup-06-event-router.sh
+++ b/files/setup-06-event-router.sh
@@ -9,6 +9,9 @@ set -euo pipefail
 echo -e "\e[92mDeploying VMware Event Router ..." > /dev/console
 kubectl --kubeconfig /root/.kube/config -n vmware create secret generic event-router-config --from-file=${EVENT_ROUTER_CONFIG}
 
+# Retrieve the version tag for VMware Event Router image
+EVENT_ROUTER_VERSION=$(awk '/Version:/ {print $2}' /etc/veba-release)
+
 cat > /root/config/event-router-k8s.yaml << __EVENT_ROUTER_CONFIG
 apiVersion: apps/v1
 kind: Deployment
@@ -27,7 +30,8 @@ spec:
         app: vmware-event-router
     spec:
       containers:
-      - image: vmware/veba-event-router:latest
+      - image: vmware/veba-event-router:${EVENT_ROUTER_VERSION}
+        imagePullPolicy: IfNotPresent
         args: ["-config", "/etc/vmware-event-router/event-router-config.json", "-verbose"]
         name: vmware-event-router
         resources:
--- a/photon.json
+++ b/photon.json
@@ -63,6 +63,9 @@
     },
     {
       "type": "shell",
+      "environment_vars": [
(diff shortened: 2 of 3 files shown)


files/setup-06-event-router.sh
```diff
@@ -9,6 +9,9 @@ set -euo pipefail
 echo -e "\e[92mDeploying VMware Event Router ..." > /dev/console
 kubectl --kubeconfig /root/.kube/config -n vmware create secret generic event-router-config --from-file=${EVENT_ROUTER_CONFIG}
 
+# Retrieve the version tag for VMware Event Router image
+EVENT_ROUTER_VERSION=$(awk '/Version:/ {print $2}' /etc/veba-release)
```
embano1, line 13: I think that's a good approach since we can always create custom (beta) builds of the appliance/router image which would be reflected in this file as well.

photon.json
```diff
@@ -63,6 +63,9 @@
     },
     {
       "type": "shell",
+      "environment_vars": [
+        "VEBA_VERSION={{ user `VEBA_VERSION` }}"
```
embano1, line 67: Where does `{{ user `VEBA_VERSION` }}` come from? Why do we need this variable and cannot reuse the one above/similar approach?

scripts/photon-containers.sh
```diff
@@ -32,7 +32,7 @@ envoyproxy/envoy:v1.11.1
 prom/prometheus:v2.11.0
 prom/alertmanager:v0.18.0
 nats-streaming:0.11.2
-vmware/veba-event-router:latest
+vmware/veba-event-router:${VEBA_VERSION}
```
embano1, line 35: Related question above.

photon.json
lamw: Good question and yes it may not look apparent :)

Let me explain. 

Since we're building the appliance from scratch, we don't have `/etc/veba-release` file at all, since there's no OS, so its sort of a chicken/egg problem and hence we can't use the method above which executes AFTER customer deploys the appliance and has the file.

In this phase, we need to grab the version from `VERSION` file which is passed in from https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/development/build.sh#L14 and then this gets exposed as an ENV variable within the section of the pre-installer within Packer. So this translate to the $VEBA_VERSION variable containing the value to then determine the VMware Event Router tag to pull (for air-gap). This all happens as part of installing the OS and setting things up.

embano1: Gotcha, if it’s coming from `VERSION` we’re good.

PR #107, opened 2020-04-10
Add invoke PagerDuty Incident sample function

pksrc:
Created a simple Pagerduty function that will make a rest API call triggering a Pagerduty Incident upon a vCenter Event. 

Function was tested on vCenter 6.7, Pagerduty(SaaS) and VEBA 0.3(w/ OpenFaaS) with VMPowerOn/Off Events. For convenience, additional unit tests are provided within the handler.py for quick local testing of the function. 

Signed-off-by: pksrc <kandasamyp@vmware.com>

### DIFF
--- a/examples/python/trigger-pagerduty-incident/README.MD
+++ b/examples/python/trigger-pagerduty-incident/README.MD
@@ -1,0 +1,98 @@
+### Get the example function
+
+Clone this repository which contains the example functions. 
+
+```bash
+git clone https://github.com/vmware-samples/vcenter-event-broker-appliance
+cd vcenter-event-broker-appliance/examples/python/trigger-pagerduty-incident
+git checkout master
+```
+
+### Customize the function
+
+For security reasons to not expose sensitive data we will create a Kubernetes [secret](https://kubernetes.io/docs/concepts/configuration/secret/) which will hold the routingkey information for the PagerDuty Event API v2. This secret will be mounted into the function during runtime and is taken care of by the appliance. We only have to create the secret with a simple command through `faas-cli`.
+
+First, change the configuration file [pdconfig.json](pdconfig.json) holding your sensitive pagerduty information in this folder:
+
+```json
+{
+    "routing_key": "<replace with your routing key>",
+    "event_action": "trigger" 
+}
+```
+
+Now go ahead and store this configuration file as secret in the appliance.
+
+```bash
+# set up faas-cli for first use
+export OPENFAAS_URL=https://VEBA_FQDN_OR_IP
+faas-cli login -p VEBA_OPENFAAS_PASSWORD --tls-no-verify # vCenter Event Broker Appliance is configured with authentication, pass in the password used during the vCenter Event Broker Appliance deployment process
+
+# now create the secret
+faas-cli secret create pdconfig --from-file=pdconfig.json --tls-no-verify
+```
+
+> **Note:** Delete the local `pdconfig.json` after you're done with this exercise to not expose this sensitive information.
+
+Lastly, define the vCenter event which will trigger this function. Such function-specific settings are performed in the `stack.yml` file. Open and edit the `stack.yml` provided with in the Python example code. Change `gateway` and `topic` as per your environment/needs.
+
+> **Note:** A key-value annotation under `topic` defines which VM event should trigger the function. A list of VM events from vCenter can be found [here](https://code.vmware.com/doc/preview?id=4206#/doc/vim.event.VmEvent.html). Multiple topics can be specified using a `","` delimiter syntax, e.g. "`topic: "VmPoweredOnEvent,VmPoweredOffEvent"`".
+
+```yaml
+provider:
+  name: openfaas
+  gateway: https://VEBA_FQDN_OR_IP # replace with your VMware Event Broker Appliance environment
+functions:
+  pdinvoke-fn:
+    lang: python3
+    handler: ./handler
+    image: vmware/veba-pagerduty-invoke:latest
+    environment:
+      write_debug: true #function writes verbose entries to the log when set to true, also requires combine_output to be set to false to avoid debug messages from showing up in the response
+      read_debug: true
+      combine_output: false #prevents error logs from showing up on the response output
+      insecure_ssl: true #set to true if you have a trusted TLS certificate on the gateway
+    secrets:
+      - pdconfig # update file with your Pagerduty integration key - https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2
+    annotations:
+      topic: VmPoweredOnEvent,VmPoweredOffEvent # or DrsVmPoweredOnEvent in a DRS-enabled cluster
+```
+
+> **Note:** If you are running a vSphere DRS-enabled cluster the topic annotation above should be `DrsVmPoweredOnEvent`. Otherwise the function would never be triggered.
+
+### Deploy the function
+
+After you've performed the steps and modifications above, you can go ahead and deploy the function:
+
+```bash
+faas-cli template pull # only required during the first deployment
+faas-cli deploy -f stack.yml --tls-no-verify
+Deployed. 202 Accepted.
+```
+
+### Trigger the function
+
+Turn on a virtual machine, e.g. in vCenter or via `govc` CLI, to trigger the function via a `(DRS)VmPoweredOnEvent`. You should now receive a PagerDuty notification through the configured notification channel
+
+> **Note:** If you are not seeing the PagerDuty alert upon the event being triggered, validate that the routing key in the `pdconfig.json` is correct and see the [troubleshooting](#troubleshooting) section below.
+
+## Troubleshooting
+
+If your PagerDuty event is not getting invoked, verify:
+
+- Routing Key in the `pdconfig.json`
+- Whether the components can talk to each other (VMware Event Router to vCenter and OpenFaaS, VMware Event Broker Appliance to PagerDuty)
+- If you have changed the `topic` in `stack.yml`, please ensure that the Function is also updated to handle the expected event data. 
+- Check the logs:
+
+```bash
+faas-cli logs pdinvoke-fn --follow --tls-no-verify 
+
+# Successful log message in the OpenFaaS function
+2019/01/25 23:48:55 Forking fprocess.
+2019/01/25 23:48:55 Query
+2019/01/25 23:48:55 Path  /
+
+{"status": "200", "message": "successfully triggered event action with dedup_key: <redacted>"}
+2019/01/25 23:48:56 Duration: 1.551482 seconds
+```
--- a/examples/python/trigger-pagerduty-incident/handler/handler.py
+++ b/examples/python/trigger-pagerduty-incident/handler/handler.py
@@ -1,0 +1,216 @@
+import sys, json, os
+import urllib3
+import requests
+import traceback
+
+# GLOBAL_VARS
+DEBUG=False
+class bgc:
+    HEADER = '\033[95m'
+    OKBLUE = '\033[94m'
+    OKGREEN = '\033[92m'
+    WARNING = '\033[93m'
+    FAIL = '\033[91m'
+    ENDC = '\033[0m'
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
+
+if(os.getenv("insecure_ssl")):
+    # Surpress SSL warnings
+    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+if(os.getenv("write_debug")):
+    sys.stderr.write(f"{bgc.WARNING}WARNING!! DEBUG has been enabled for this function. Sensitive information could be printed to sysout{bgc.ENDC} \n")
+    DEBUG=True
+
+def debug(s):
+    if DEBUG:
+        sys.stderr.write(s+" \n") #syserr only get logged on the console logs
+
+#
+### Paths and Endpoints
+### More examples and details here - https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2
+#
+PAGERDUTY_API_PATH='https://events.pagerduty.com/v2/enqueue'
+PD_CONFIG='/var/openfaas/secrets/pdconfig'
+class FaaSResponse:
+    """
+    FaaSResponse is a helper class to construct a properly formatted message returned by this function.
+    By default, OpenFaaS will marshal this response message as JSON.
+    """    
+    def __init__(self, status, message):
+        """
+        Arguments:
+            status {str} -- the response status code
+            message {str} -- the response message
+        """        
+        self.status=status
+        self.message=message
+
+class Pagerduty:
+    """
+    Pagerduty is a Class used to trigger Events in PagerDuty.
+    """    
+
+    def __init__(self,conn):
+        """
+        Arguments:
+            conn {session} -- connection to PagerDuty REST API
+        """
+        self.session=conn
+
+    # PagerDuty REST API implementation        
+    def invoke(self,obj):
+        """
+        Make a rest api call to Pagerduty Events API
+        
+        Arguments:
+            obj {dict} -- Generated API Body for the PagerDuty Events API
+
+        Returns:
+            FaaSResponse -- status code and message
+        """
+        try:
+            resp = self.session.post(PAGERDUTY_API_PATH,json=obj)
+            resp.raise_for_status()
+            debug(f'{bgc.OKGREEN}HTTP POST Request successful{bgc.ENDC}')
+            debug(f'{bgc.OKBLUE}Response Body > {bgc.ENDC}{resp.text}')
+            resp_body = json.loads(resp.text)
+            return FaaSResponse('200', 'Successfully invoked PagerDuty API! dedup_key for this request: {0}'.format(resp_body['dedup_key']))
+        except requests.HTTPError as err:
+            return FaaSResponse('500', 'Could not invoke PagerDuty API > HTTPError: {0}'.format(err))
+
+def handle(req):
+    
+    # Validate Event input
+    debug(f'{bgc.HEADER}---Validating CloudEvent--- {bgc.ENDC}')
+    debug(f'{bgc.OKBLUE}Event (raw) > {bgc.ENDC}{req}')
+    try:
+        cevent = json.loads(req)
+        debug(f'{bgc.OKGREEN}Successfully parsed Event into JSON!{bgc.ENDC}')
+        debug(f'{bgc.OKBLUE}Event (JSON) > {bgc.ENDC}{json.dumps(cevent, indent=4, sort_keys=True)}')
+    except json.JSONDecodeError as err:
+        res = FaaSResponse('400','Invalid JSON > JSONDecodeError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return
+    
+    # Validate Config file
+    debug(f'{bgc.HEADER}---Validating Config--- {bgc.ENDC}')
+    debug(f'{bgc.OKBLUE}Reading Config File > {bgc.ENDC}{PD_CONFIG}')
+    try: 
+        with open(PD_CONFIG, 'r') as pdconfigfile:
+            pdconfig = json.load(pdconfigfile)
+            debug(f'{bgc.OKGREEN}Successfully parsed Configuration into JSON!{bgc.ENDC}')
+            debug(f'{bgc.OKBLUE}Configuration > {bgc.ENDC}{json.dumps(pdconfig,indent=4)}')
+            routingkey=pdconfig['routing_key']
+            event_action=pdconfig['event_action']
+    except json.JSONDecodeError as err:
+        res = FaaSResponse('400','Invalid JSON > JSONDecodeError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return
+    except KeyError as err:
+        res = FaaSResponse('400','Required key not found in the provided configuration > KeyError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return 
+    except OSError as err:
+        res = FaaSResponse('500','Could not read PagerDuty configuration > OSError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return
+
+    # Assert that the function is able to get the required information from the event and build the request body
+    # For debugging: validate the JSON blob we received - uncomment print statements if needed
+    # print(cevent)
+    debug(f'{bgc.HEADER}---Building HTTP Request body--- {bgc.ENDC}')
+    try:
+        # Map the CloudEvent data and build the PagerDuty Event API Request body
+        obj = {
+                'routing_key': routingkey,
+                'event_action': event_action,
+                'client': 'VMware Event Broker Appliance',
+                'client_url': cevent['source'],
+                'payload': {
+                    'summary': cevent['data']['FullFormattedMessage'],
+                    'timestamp': cevent['data']['CreatedTime'],
+                    'source': cevent['source'],
+                    'severity': 'info',
+                    'component': cevent['data']['Vm']['Name'],
+                    'group': cevent['data']['Host']['Name'],
+                    'class': cevent['subject'],
+                    'custom_details': {
+                        'user': cevent['data']['UserName'],
+                        'Datacenter': cevent['data']['Datacenter'],
+                        'ComputeResource': cevent['data']['ComputeResource'],
+                        'Host': cevent['data']['Host'],
+                        'VM': cevent['data']['Vm']
+                    }
+                }
+                #,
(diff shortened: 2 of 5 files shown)


examples/python/trigger-pagerduty-incident/README.MD
```diff
@@ -0,0 +1,95 @@
...
+```bash
+# set up faas-cli for first use
+export OPENFAAS_URL=https://VEBA_FQDN_OR_IP
+faas-cli login -p VEBA_OPENFAAS_PASSWORD --tls-no-verify # vCenter Event Broker Appliance is configured with authentication, pass in the password used during the vCenter Event Broker Appliance deployment process
+
+# now create the secret
+faas-cli secret create pdconfig --from-file=pdconfig.json --tls-no-verify
+```
+
+**Note:** Delete the local `pdconfig.json` after you're done with this exercise to not expose this sensitive information.
```
embano1, line 35: Can we make these consistent with 

```
> **Note:** Lorem ipsum...
```

Also see https://github.com/vmware-samples/vcenter-event-broker-appliance/pull/108

examples/python/trigger-pagerduty-incident/README.MD
```diff
@@ -0,0 +1,95 @@
...
+faas-cli template pull # only required during the first deployment
+faas-cli deploy -f stack.yml --tls-no-verify
+Deployed. 202 Accepted.
+```
+
+### Trigger the function
+
+Turn on a virtual machine, e.g. in vCenter or via `govc` CLI, to trigger the function via a `(DRS)VmPoweredOnEvent`. You should now receive a PagerDuty notification through the configured notification channel
+
+**Note:** If you are not seeing the PagerDuty alert upon the event bring triggered, validate that the routing key in the `pdconfig.json` is correct and see the [troubleshooting](#troubleshooting) section below.
```
embano1, line 75: Typo "bring"

examples/python/trigger-pagerduty-incident/README.MD
```diff
@@ -0,0 +1,95 @@
...
+
+## Troubleshooting
+
+If your PagerDuty event is not getting invoked, verify:
+
+- Routing Key in the `pdconfig.json`
+- Whether the components can talk to each other (VMware Event Router to vCenter and OpenFaaS, function to PagerDuty)
+- Check the logs (`kubectl` is installed and configured locally on the appliance)):
```
embano1, line 83: Example below uses faas-cli not kubectl

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
+import sys, json, os
+import urllib3
+import requests
+import json
+
+# Surpress SSL warnings
```
embano1, line 6: Can we make this optional, e.g. via environment var, please? We should not default to insecure in our functions (older ones likely need cleanup as well...)

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+
+class Pagerduty:
+    """Pagerduty is a Class used to trigger Events in PagerDuty."""    
+
+    def __init__(self,conn):
+        """
+        
+        Arguments:
+
+            conn {Session} -- connection to vCenter REST API
```
embano1, line 39: "vCenter..."?

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+        """        
+
+        try:
+             with open(PD_CONFIG, 'r') as pdconfigfile:
+                pdconfig = json.load(pdconfigfile)
+                self.routingkey=pdconfig['routing_key']
+                self.event_action=pdconfig['event_action']
+        except OSError as e:
+            print(f'could not read PagerDuty configuration: {e}')
+            sys.exit(1)
```
embano1, line 49: Can we return (with error logging) here instead of sys.exit? Failures should be handled gracefully.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+        except KeyError as e:
+            print(f'mandatory configuration key not found: {e}')
+            sys.exit(1)
```
embano1, line 52: See above.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+            'event_action': 'REPLACE',
+            'client': 'vCenter',
+            'client_url': ref['source']
+        }
+
+    #print(obj)
+
+    # Make the Rest Api Call to PagerDuty
+    s=requests.Session()
+    s.verify=False
```
embano1, line 133: See above regarding defaulting to insecure

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+        return
+
+    # Map the CloudEvent data and build the PagerDuty Event API Request body
+    obj = {
+            'payload': {
+                'summary': eventdata['FullFormattedMessage'],
+                'timestamp': eventdata['CreatedTime'],
+                'source': ref['source'],
+                'severity': 'info',
+                'component': eventdata['Vm']['Name'],
```
embano1, line 103: What if this key does not exist, e.g. someone subscribed to the wrong event. It's probably a good idea to check for these fields (Vm, Host) in the event payload received with try/except and handle incorrect gracefully (with appropriate error message).

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+                'src': 'https://www.pagerduty.com/wp-content/uploads/2016/05/pagerduty-logo-green.png',
+                'href': 'https://example.com/',
+                'alt': 'Image of VM'
+            }],
+            'links': [{
+                'href': 'https://example.com/',
+                'text': 'Link to VM'
+            }],
+            'event_action': 'REPLACE',
+            'client': 'vCenter',
```
embano1, line 125: To discuss: should the client be VEBA for better troubleshooting? vCenter source is already covered in the payload earlier.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,143 @@
...
+    # Make the Rest Api Call to PagerDuty
+    s=requests.Session()
+    s.verify=False
+    pg = Pagerduty(s)
+    res = pg.connect(obj)
+    if res.status != '200':
+        print(json.dumps(vars(res)))
+        return
+
+    # Close session to VC
```
embano1, line 140: != VC

examples/python/trigger-pagerduty-incident/stack.yml
```diff
@@ -0,0 +1,15 @@
+provider:
+  name: openfaas
+  gateway: https://127.0.0.1:31112
```
embano1, line 3: Please see other examples how we express gateway in examples.

examples/python/trigger-pagerduty-incident/stack.yml
```diff
@@ -0,0 +1,15 @@
...
+functions:
+  pdinvoke-fn:
+    lang: python3
+    handler: ./handler
+    image: pkbu/veba-pagerduty-invoke:latest
```
embano1, line 8: Please replace with official VEBA VMware image

examples/python/trigger-pagerduty-incident/stack.yml
```diff
@@ -0,0 +1,15 @@
...
+    environment:
+      write_debug: true
+      read_debug: true
+    secrets:
+      - pdconfig
+    annotations:
+      topic: VmPoweredOnEvent, VmPoweredOffEvent
```
embano1, line 15: Please remove the whitespace after the ","

examples/python/trigger-pagerduty-incident/stack.yml
pksrc: do i then build and push to the vmware repository?

examples/python/trigger-pagerduty-incident/stack.yml
pksrc: curious, is this change requested as a yaml/k8s best practice?

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: for the api body w/
```
  "client": "vCenter",
  "client_url": "https://10.0.10.90/sdk",
```

here is an example of how the client value gets rendered within PagerDuty
image

Agreed on the client being VEBA but the question is around the URL, what should we be using for the client URL?

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: https://code.vmware.com/apis/704/vsphere/vim.event.Event.html - Does this mean that there are some event attributes that are generated with every event? Can i skip validation for the required fields noted for the Event Object?

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: done

examples/python/trigger-pagerduty-incident/stack.yml
pksrc: I've updated the readme accordingly but i don't think i have permissions to push my docker build to the vmware repo. I'm leaving this pointing to my image at this time and have you guys update this for me once the image is made available from vmware repo. fair?

pksrc:
@embano1 I've update per your recommendation. There were a few outstanding questions that i had (commented on the conversation).

examples/python/trigger-pagerduty-incident/README.MD
```diff
@@ -0,0 +1,96 @@
...
+  name: openfaas
+  gateway: https://VEBA_FQDN_OR_IP # replace with your vCenter Event Broker Appliance environment
+functions:
+  pdinvoke-fn:
+    lang: python3
+    handler: ./handler
+    image: vmware/veba-pagerduty-invoke:latest
+    environment:
+      write_debug: true
+      read_debug: true
```
embano1, line 52: You might want to add the `insecure_ssl` option here as well so it's apparent to the user

examples/python/trigger-pagerduty-incident/stack.yml
embano1: tbh, not sure how YAML interprets spaces in strings - perhaps it just passes them through as is. and in this case it's up to the library to interpret. would have to look up how the openfaas controller SDK handles whitespaces, but I think it's always safer to not assume correct whitespace handling and thus remove the space(s)

examples/python/trigger-pagerduty-incident/README.MD
```diff
@@ -0,0 +1,96 @@
...
+    secrets:
+      - pdconfig # update this with your Pagerduty integration key - https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2
+    annotations:
+      topic: VmPoweredOnEvent, VmPoweredOffEvent # or DrsVmPoweredOnEvent in a DRS-enabled cluster
```
embano1, line 56: As discussed, please remove the whitespace here and in the `stack.yaml`

examples/python/trigger-pagerduty-incident/stack.yml
embano1: For local testing, use your image. In this PR please replace this line with an official image from VMware. Once the PR is merged, we'll push the image to the VMware hub.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+                #    'href': 'https://example.com/',
+                #    'alt': eventdata['Vm']
+                # }],
+                # 'links': [{
+                #    'href': 'https://example.com/',
+                #    'text': 'Link to VM'
+                # }]
+            }
+    except KeyError as err:
+        res = FaaSResponse('400','JSON does not conform to CloudEvent spec > KeyError: {0}'.format(err))
```
embano1, line 125: This error message is strictly speaking not correct. If you fail to look up a key, e.g. `eventdata['Host']['Name']` in an event which has this set to`NULL`, the event still conforms to the CloudEvent spec. I think it's better to change the response message to something like "required key not found in JSON...".

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+    except ValueError as err:
+        res = FaaSResponse('400','invalid JSON > ValueError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return
+    
+    # Assert that the Event conforms to CloudEvent Spec
+    # For debugging: validate the JSON blob we received - uncomment print statements if needed
+    # print(cevent)
+    try:
+        eventdata = (cevent['data'])
```
embano1, line 90: Is this required and what does it do? (use of tuple)

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+                # 'links': [{
+                #    'href': 'https://example.com/',
+                #    'text': 'Link to VM'
+                # }]
+            }
+    except KeyError as err:
+        res = FaaSResponse('400','JSON does not conform to CloudEvent spec > KeyError: {0}'.format(err))
+        print(json.dumps(vars(res)))
+        return
+    except TypeError as err:
```
embano1, line 128: Not sure if exception `TypeError` really helps here. What is this supposed to catch?

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+        s=requests.Session()
+        if(os.getenv("insecure_ssl")):
+            s.verify=False
+        pg = Pagerduty(s)
+        res = pg.invoke(obj)
+        if res.status != '200':
+            print(json.dumps(vars(res)))
+    except (OSError, KeyError) as err:
+        # this could only occur from the PagerDuty constructor and sufficient logging for these Exceptions are handled there. 
+        return
```
embano1, line 146: Since you're handling the exception in the constructor, why not return a proper error message like with `pg.invoke` above and avoid nested exception handling? I.e. handle exceptions in the constructor/invocation and return proper FaaSResponse in this handler? Avoids jumping around in the code to see what's going on, less ambiguity and more clarity to the reader.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+            s.verify=False
+        pg = Pagerduty(s)
+        res = pg.invoke(obj)
+        if res.status != '200':
+            print(json.dumps(vars(res)))
+    except (OSError, KeyError) as err:
+        # this could only occur from the PagerDuty constructor and sufficient logging for these Exceptions are handled there. 
+        return 
+    except Exception as err:
+        res = FaaSResponse('400','Unexpected Error occured > Exception: {0}'.format(err))
```
embano1, line 148: Unexpected/internal server errors must be 5xx not 4xx. Can you please review those sections and change accordingly?

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        try:
+             with open(PD_CONFIG, 'r') as pdconfigfile:
+                pdconfig = json.load(pdconfigfile)
+                self.routingkey=pdconfig['routing_key']
+                self.event_action=pdconfig['event_action']
+        except OSError as err:
+            res = FaaSResponse('400','Could not read PagerDuty configuration > OSError: {0}'.format(err))
+            print(json.dumps(vars(res)))
+            raise
```
embano1, line 48: See comment below on how I think this could be simplified, avoiding nested try/catch and (re)raising.

embano1:
> @embano1 I've update per your recommendation. There were a few outstanding questions that i had (commented on the conversation).

Thx, left some notes/comments as well.

Regarding the client_url/referral I think it's fine for now to comment this section out as you did. If you want to leave it in I have no preference and I'm good with whatever you decide.

examples/python/trigger-pagerduty-incident/stack.yml
pksrc: done

examples/python/trigger-pagerduty-incident/stack.yml
pksrc: done

examples/python/trigger-pagerduty-incident/README.MD
pksrc: done

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: Updated the exception handling to be more accurate and avoided nested/complex flows

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: updated. didn't know (variable) had a special meaning in python

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: this is required to catch null values where i expect dicts such as `'component': cevent['data']['Vm']['Name'],` and vm can be set to null with some events.

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: Updated the exception handling to be more accurate and avoided nested/complex flows

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: updated. thanks for the catch

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: Updated the exception handling to be more accurate and avoided nested/complex flows

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -1,11 +1,30 @@
...
 if(os.getenv("insecure_ssl")):
+    # Surpress SSL warnings
     urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+if(os.getenv("write_debug")):
+    sys.stderr.write(f"{bgc.WARNING}WARNING!! DEBUG has been enabled for this function. Sensitive information could be printed to sysout{bgc.ENDC} \n ")
+    DEBUG=True
+
+def debug(s):
+    if DEBUG:
+        sys.stderr.write(s+" \n ") #Syserr only get logged on the console logs
```
embano1, line 27: Did you verify this just locally or also in a fully working VEBA environment? Asking because I'm not sure how this is interpreted in Kubernetes/OpenFaaS.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -63,94 +69,112 @@ def invoke(self,obj):
...
-        obj['routing_key'] = self.routingkey
-        obj['event_action'] = self.event_action
         try:
             resp = self.session.post(PAGERDUTY_API_PATH,json=obj)
             resp.raise_for_status()
-            print(resp.text)
+            debug(f'{bgc.OKGREEN}Response Body > {bgc.ENDC}{resp.text}')
             resp_body = json.loads(resp.text)
-            return FaaSResponse('200', 'Successfully {0}ed event action with dedup_key: {1}'.format(self.event_action, resp_body['dedup_key']))
+            return FaaSResponse('200', 'Successfully triggered event action with dedup_key: {0}'.format(resp_body['dedup_key']))
```
embano1, line 77: Can we align this with the error message, e.g. 'Successfully invoked Pagerduty API...' so it's less ambiguous?

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -63,94 +69,112 @@ def invoke(self,obj):
...
 def handle(req):
     # Validate Event input
     try:
         cevent = json.loads(req)
-    except ValueError as err:
-        res = FaaSResponse('400','invalid JSON > ValueError: {0}'.format(err))
+    except json.decoder.JSONDecodeError as err:
+        res = FaaSResponse('400','Invalid JSON > JSONDecodeError: {0}'.format(err))
         print(json.dumps(vars(res)))
+        debug(f'{bgc.OKBLUE}Event > {bgc.ENDC}{req}')
```
embano1, line 88: Same question as above regarding OpenFaaS/Kubernetes interpretation of stderr output. Whatever is printed here, OpenFaaS will only interpret this section `print(json.dumps(vars(res)))` to indicate success/failure? I'd rather inverse the logic, i.e. print debug before the final `print` and `return`.

examples/python/trigger-pagerduty-incident/handler/handler.py
```diff
@@ -63,94 +69,112 @@ def invoke(self,obj):
...
     
-    # Assert that the Event conforms to CloudEvent Spec
+    # Validate Config file
+    try: 
+        with open(PD_CONFIG, 'r') as pdconfigfile:
+            pdconfig = json.load(pdconfigfile)
+            routingkey=pdconfig['routing_key']
+            event_action=pdconfig['event_action']
+    except OSError as err:
+        res = FaaSResponse('400','Could not read PagerDuty configuration > OSError: {0}'.format(err))
```
embano1, line 98: One could argue that this and the next error should be 500s because they relate to internal server logic and not the provided payload (event data) in this function.

embano1:
After we passed the review, can you please rebase and squash the commits into one and reword the commit message to `Add PagerDuty Python Example`?

Please also describe in the PR how you tested this function (environment, version, etc).

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: this was verified on OpenFaaS (VEBA v0.3)

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: OpenFaaS by default combines the stdout and stderr to the http response. However, there is a env variable `combine_output: false` that prevents the stderr from showing up on http response back to OpenFaaS and is only logged on the console. 

I have it ordered this way to help show the error first and then supporting data below the error for context.

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: I am a little conflicted on treating these functions as a client or server.  I'll have to research the appropriate standard if there is one.

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: done

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: Based on the understanding that i have and from research, a 400 error is something that could be fixed by the caller/function user i.e send the function the right event and the correct config and function should work. A 500 error is something that could only be fixed by the function owner or occurs for other reasons that are not under the user's control i.e resource constraints, n/w errors or unexpected exceptions. 

Let me know if you have a different viewpoint.

pksrc:
@embano1 updated the succes message and left comments on the rest of the changes requested. I've also rebased and pushed as a single commit. Let me know if i've missed anything.

examples/python/trigger-pagerduty-incident/handler/handler.py
embano1: > A 500 error is something that could only be fixed by the function owner

Exactly, that's why I think the error on line 98 should be a `500`.

examples/python/trigger-pagerduty-incident/handler/handler.py
pksrc: I've  reordered and updated the debug messaging now.

embano1:
Thank you @pksrc 
Can you please squash into one commit also describing how you tested this function? Then we're good to merge.

pksrc:
Thanks @embano1 for your thorough review and appreciate your patience :) . I've squashed into one commit with updated commit message.

embano1:
Thank you, can you please update the PR (first comment) with details how you tested the function?

PR #110, opened 2020-04-15
Add Function for Rest API Integrations (Slack, Pagerduty, JIRA, ServiceDesk, ZenDesk and ServiceNow)

pksrc:
This function is the Postman or cURL for VEBA and can make a POST api call to any endpoint that supports basic, token auth in headers or no authentication. 
Sample configs provided for above mentioned endpoints
Debug messages print to stderr, does not show in FaaS Response. 
Unit tests available in handle.py, uncomment to test locally

**Testing**: 
Environment: vCenter 6.7 and VEBA v0.3(w/ OpenFaaS)
This function was tested against the following APIs/Systems (configs provided
- Slack (send message to a channel)
- PagerDuty (trigger an incident) and 
- ServiceNow (create an incident) 
- ServiceDesk (create a request)
- JIRA (create a JIRA issue)
- Zendesk (create a ticket) 

Signed-off-by: pksrc <kandasamyp@vmware.com>

### DIFF
--- a/examples/python/invoke-rest-api/README.MD
+++ b/examples/python/invoke-rest-api/README.MD
@@ -1,0 +1,199 @@
+# Function to make a POST request to any REST API
+
+This function aims to provide a easy way to make HTTP Post request to an API Endpoint that might help with a lot of integration scenarios. Eg. Post to a Slack channel, Create a PagerDuty or ServiceNow incident. 
+
+> **NOTE:** This function currently supports endpoints that allow basic authentication (un/pwd) or token based authentication that can be passed with the headers
+
+- [Deploy](#deploy)
+  * [Get the example function](#get-the-example-function)
+  * [Customize the function](#customize-the-function)
+    + [Understanding the Metaconfig-[SYSTEM].json](#understanding-the-metaconfig--system-json)
+      - [Provide the API Details](#provide-the-api-details)
+      - [Mapping the Events and Request body](#mapping-the-events-and-request-body)
+    + [Updating the Stack.yml](#updating-the-stackyml)
+    + [Updating the Handler.py (advanced)](#updating-the-handlerpy--advanced-)
+  * [Deploy the function](#deploy-the-function)
+    + [Create the secret](#create-the-secret)
+    + [Build function (only if handler.py is changed)](#build-function--only-if-handlerpy-is-changed-)
+    + [Deploy the function](#deploy-the-function-1)
+  * [Trigger the function](#trigger-the-function)
+- [Troubleshooting](#troubleshooting)
+
+# Deploy
+
+## Get the example function
+
+Clone this repository which contains the example functions. 
+
+```bash
+git clone https://github.com/vmware-samples/vcenter-event-broker-appliance
+cd vcenter-event-broker-appliance/examples/python/invoke-rest-api
+git checkout master
+```
+
+## Customize the function
+
+There are three key files, that you might have to modify if you are looking to customize this function and make a post api call to an external system. 
+
+```bash
+  /invoke-rest-api/metaconfig-[SYSTEM].json #sample copies provided for PagerDuty, Slack, JIRA, Zendesk, ServiceDesk and ServiceNow
+  /invoke-rest-api/stack.yml 
+  /invoke-rest-api/handler/handler.py
+```
+
+### Understanding the Metaconfig-[SYSTEM].json 
+First, change the configuration file `metaconfig-[SYSTEM].json` (samples for Slack, PagerDuty and ServiceNow are provided with this example) holding both sensitive and configuration information needed for this function in this folder. 
+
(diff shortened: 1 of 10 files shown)


pksrc:
@embano1 - if you can also take a look at this function when you get a chance? Similarly structured to the pagerduty function but more capable for making a post request to any rest api endpoint. Let me know if there are any changes that need to be made.

examples/python/invoke-rest-api/README.MD
```diff
@@ -0,0 +1,135 @@
...
+
+```bash
+git clone https://github.com/vmware-samples/vcenter-event-broker-appliance
+cd vcenter-event-broker-appliance/examples/python/invoke-rest-api
+git checkout master
+```
+
+### Customize the function
+
+There are three key files, that you might have to modify if you are looking to customize this function and make a post api call to an external system.
```
embano1, line 17: Can you please also describe the case for modifying the other two files (stack/handler) or when that would be required? So it's easier to follow for newcomers.

examples/python/invoke-rest-api/stack.yml
```diff
@@ -0,0 +1,17 @@
+provider:
+  name: openfaas
+  gateway: https://VEBA_FQDN_OR_IP
+functions:
+  restpost-fn:
+    lang: python3
+    handler: ./handler
+    image: vmware/veba-rest-post:latest
```
embano1, line 8: Can you please align this with the function name?

examples/python/invoke-rest-api/README.MD
```diff
@@ -0,0 +1,135 @@
...
+
+Now let's store this configuration file as secret in the appliance.
+
+```bash
+# set up faas-cli for first use
+export OPENFAAS_URL=https://VEBA_FQDN_OR_IP
+faas-cli login -p VEBA_OPENFAAS_PASSWORD --tls-no-verify # vCenter Event Broker Appliance is configured with authentication, pass in the password used during the vCenter Event Broker Appliance deployment process
+
+# now create the secret
+faas-cli secret create metaconfig --from-file=metaconfig-pduty.json --tls-no-verify
```
embano1, line 66: Just double-checking here: is the actual file name of the secret file being projected into the function then `metaconfig` or `metaconfig-pduty.json`? (Because the function uses not the latter)

examples/python/invoke-rest-api/README.MD
```diff
@@ -0,0 +1,135 @@
...
+```
+
+> **Note:** Delete the local `metaconfig.json` after you're done with this exercise to not expose any sensitive information.
```
embano1, line 69: I see inconsistent use of the `.json` file names across the readme. Can you please standardize on something like `metaconfig-[REST-API].json` so it matches the syntax of your examples provided?

examples/python/invoke-rest-api/metaconfig-pduty.json
```diff
@@ -0,0 +1,60 @@
+{
+    "url": "https://events.pagerduty.com/v2/enqueue",
+    "headers": {
+        "content-type": "application/json; charset=UTF-8"
+    },
+    "auth": {},
```
embano1, line 6: Is it assumed that users understand if and what to provide respective to authentication for each of these metaconfigs? I cannot find anything in the README.

examples/python/invoke-rest-api/README.MD
pksrc: I've kept the secret name generic to be `metaconfig` which the function is looking to read. 

With one set of users, i'm anticipating that there won't be a need to change the function and  they will just end up providing the right `metaconfig` secret (for the api info and the event info) and deploy with the container image quickly. 

For the other set of users who want to have multiple copies of this function for different systems, they'll need to have multiple copies of the metaconfig-[SYSTEM].json created as secret and also modify the `handler.py` with the appropriate config file path to the secret.

examples/python/invoke-rest-api/README.MD
pksrc: good idea, i'll use metaconfig-[SYSTEM].json

examples/python/invoke-rest-api/metaconfig-pduty.json
pksrc: Yes, most apis provide some sort of api key to use and don't seem to require an auth. With some others, they might require basic authentication which can be send in the header. 

Where the apis require username and password explicitly, i've added that as an option in the config. I only provide limited support in terms of the possible auth options. I'll document this in the readme so there is no ambiguity.

examples/python/invoke-rest-api/stack.yml
pksrc: done

examples/python/invoke-rest-api/README.MD
pksrc: will do, this will be a long readme :)

pksrc:
@embano1 - Updated the ReadMe and the Image in stack.yml. Also updated the PR with testing information.

embano1:
Please squash, thx!

Issue #16, opened 2019-12-10
Evaluate using Antrea for networking


embano1:
https://github.com/vmware-tanzu/antrea

cc/ @yasensim

PR #114, opened 2020-04-30
Replace Weave with Antrea CNI 

lamw:
> Summary

This changes contains the configuration updates to support Antrea as the default CNI instead of Weave. v0.6.0 was used as it contained a fix which was identified during the research phase. 

> Resolves

Closes: https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/16

> Testing

Deployed VEBA w/OpenFaa Processor against vSphere 7.0 build and verified deployment was successful and all pods were running as expected

Screen Shot 2020-04-30 at 2 49 37 PM

### DIFF
--- a/files/setup-01-os.sh
+++ b/files/setup-01-os.sh
@@ -21,3 +21,7 @@ if [ "${DOCKER_NETWORK_CIDR}" != "172.17.0.1/16" ]; then
 EOF
 systemctl restart docker
 fi
+
+echo -e "\e[92mConfiguring IP Tables for Antrea ..." > /dev/console
+iptables -A INPUT -i gw0 -j ACCEPT
+iptables-save > /etc/systemd/scripts/ip4save
--- a/files/setup-04-kubernetes.sh
+++ b/files/setup-04-kubernetes.sh
@@ -15,24 +15,30 @@ echo -e "\e[92mDisabling/Stopping IP Tables  ..." > /dev/console
 systemctl stop iptables
 systemctl disable iptables
 
+# Customize the POD CIDR Network if provided or else default to 10.10.0.0/16
+if [ -z "${POD_NETWORK_CIDR}" ]; then
+    POD_NETWORK_CIDR="10.16.0.0/16"
+fi
+
+cat >> /root/config/kubeconfig.yml << EOF
+
+networking:
+  podSubnet: ${POD_NETWORK_CIDR}
+EOF
+
 # Setup k8s
 echo -e "\e[92mSetting up k8s ..." > /dev/console
+
+echo -e "\e[92mDeloying kubeadm ..." > /dev/console
 HOME=/root
 kubeadm init --ignore-preflight-errors SystemVerification --skip-token-print --config /root/config/kubeconfig.yml
 mkdir -p $HOME/.kube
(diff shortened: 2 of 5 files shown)


scripts/photon-settings.sh
```diff
@@ -13,6 +13,7 @@ echo "net.ipv6.conf.all.disable_ipv6 = 1" >> /etc/sysctl.conf
 
 echo '> Applying latest Updates...'
 tdnf -y update
+tdnf upgrade linux-esx
```
embano1, line 16: Any packages pulled in which are changing/affecting our packages used for VEBA? (Just curious about potential regression holes)

scripts/photon-containers.sh
```diff
@@ -19,8 +12,7 @@ k8s.gcr.io/kube-proxy:v1.14.9
 k8s.gcr.io/pause:3.1
 k8s.gcr.io/etcd:3.3.10
 k8s.gcr.io/coredns:1.3.1
-docker.io/weaveworks/weave-kube:2.6.2
-docker.io/weaveworks/weave-npc:2.6.2
+antrea/antrea-ubuntu:v0.6.0
```
embano1, line 15: Assuming you also tested without network connectivity? If not, did you see the local image being used, i.e. no pull?

files/setup-01-os.sh
```diff
@@ -21,3 +21,7 @@ if [ "${DOCKER_NETWORK_CIDR}" != "172.17.0.1/16" ]; then
 EOF
 systemctl restart docker
 fi
+
+echo -e "\e[92mConfiguring IP Tables for Antrea ..." > /dev/console
+iptables -A INPUT -i gw0 -j ACCEPT
```
embano1, line 26: What is `gw0` interface? Something Antrea uses internally or exposed to the outside?

files/setup-04-kubernetes.sh
```diff
@@ -15,25 +15,31 @@ echo -e "\e[92mDisabling/Stopping IP Tables  ..." > /dev/console
 systemctl stop iptables
 systemctl disable iptables
 
+# Customize the POD CIDR Network if provided or else default to 10.10.0.0/16
```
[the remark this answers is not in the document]
lamw: This was a recommended safe default from Antrea folks, given our previous network overlaps. If there are issues, this is one of the configurable parameters within the VEBA OVF property (as it was before).

scripts/photon-containers.sh
lamw: Yup and Antrea YAML includes `IfNotPresent` policy

files/setup-01-os.sh
lamw: Yup, its an Antrea internal interface and FW rule is required

Per [Antrea Architecture](https://github.com/vmware-tanzu/antrea/blob/master/docs/architecture.md)
> On the OVS bridge, Antrea Agent also creates an internal port - gw0 by default - to be the gateway of the Node's subnet, and a tunnel port tun0 which is for creating overlay tunnels to other Nodes.

scripts/photon-settings.sh
lamw: This was recommended per Antrea folks or else we ran into an issue. Given basic functionality has been tested in VEBA, I don't believe this has caused any regression

PR #121, opened 2020-05-06
Replace static VEBA text with dynamic DCUI interface

lamw:
> Summary

Today, when VEBA completes its initial configuration, entries are appended into `/etc/issue` which provides end users with URLs based on their deployment (e.g. hostname) that they can access on VEBA including /status, /boostrap, stats and OpenFaaS URL (if that was configured as a processor). Since this file is only generated once, changes to VEBA networking such as hostname would not be reflected. 

This change replaces the static `/etc/issue` entry with a dynamic Direct Console User Interface (DUCI) application that provides the equivalent functionality. Similiar to /etc/issue, this application will automatically startup and run in tty1. Users will be able to quickly glance at the information and if they need to login, simply switching to tty2 (ALT+F2) will enable the standard console login. 

In addition, the VEBA version and commit ID are clearly visible in the DCUI interface which can enable ease of troubleshooting/debugging. 

> Resolves

N/A

> Testing

Deploy VEBA Appliance and confirmed that VEBA DCUI is automatically running

Screen Shot 2020-05-06 at 3 07 15 PM


embano1:
LGTM. One final question: systemd will always restart the process, so escaping with „Q“ essentially becomes a no-op for users?

lamw:
> LGTM. One final question: systemd will always restart the process, so escaping with „Q“ essentially becomes a no-op for users?

That's correct

PR #139, opened 2020-05-13
Allow use of "http://" prefix in HTTPS Proxy config

dsbibby:
If a http(s):// prefix is specified for either HTTP or HTTPS Proxy OVF properties, then use that rather than assuming either.

### DIFF
--- a/files/setup-02-proxy.sh
+++ b/files/setup-02-proxy.sh
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
     fi
 
     if [ ! -z "${HTTP_PROXY}" ]; then
-        if [ $YES_CREDS -eq 1 ]; then
-            HTTP_PROXY_URL="http://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY}"
-        else
-            HTTP_PROXY_URL="http://${HTTP_PROXY}"
-        fi
-        echo "HTTP_PROXY=\"${HTTP_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/http-proxy.conf << __HTTP_DOCKER_PROXY__
+        a=($(printf '%s\n' "${HTTP_PROXY//\:\/\//$'\n'}"))
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTP_PROXY_PROTOCOL=${a[0]}
+            HTTP_PROXY_SERVER_PORT=${a[1]}
+            if [ $YES_CREDS -eq 1 ]; then
+                HTTP_PROXY_URL="${HTTP_PROXY_PROTOCOL}://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY_SERVER_PORT}"
+            else
+                HTTP_PROXY_URL="${HTTP_PROXY_PROTOCOL}://${HTTP_PROXY_SERVER_PORT}"
+            fi
+            echo "HTTP_PROXY=\"${HTTP_PROXY_URL}\"" >> ${PROXY_CONF}
+            cat > ${DOCKER_PROXY}/http-proxy.conf << __HTTP_DOCKER_PROXY__
 [Service]
 Environment="HTTP_PROXY=${HTTP_PROXY_URL}" "NO_PROXY=${NO_PROXY}"
 __HTTP_DOCKER_PROXY__
+        else
+	    echo -e "\e[91mInvalid HTTP Proxy URL supplied" > /dev/console
+        fi
     fi
 
     if [ ! -z "${HTTPS_PROXY}" ]; then
-        if [ $YES_CREDS -eq 1 ]; then
-            HTTPS_PROXY_URL="https://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTPS_PROXY}"
-        else
-            HTTPS_PROXY_URL="https://${HTTPS_PROXY}"
-        fi
-        echo "HTTPS_PROXY=\"${HTTPS_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/https-proxy.conf << __HTTPS_DOCKER_PROXY__
+        a=($(printf '%s\n' "${HTTPS_PROXY//\:\/\//$'\n'}"))
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTPS_PROXY_PROTOCOL=${a[0]}
+            HTTPS_PROXY_SERVER_PORT=${a[1]}
+            if [ $YES_CREDS -eq 1 ]; then
+                HTTPS_PROXY_URL="${HTTPS_PROXY_PROTOCOL}://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTPS_PROXY_SERVER_PORT}"
+            else
+                HTTPS_PROXY_URL="${HTTPS_PROXY_PROTOCOL}://${HTTPS_PROXY_SERVER_PORT}"
+            fi
+            echo "HTTPS_PROXY=\"${HTTPS_PROXY_URL}\"" >> ${PROXY_CONF}
+            cat > ${DOCKER_PROXY}/https-proxy.conf << __HTTPS_DOCKER_PROXY__
 [Service]
 Environment="HTTPS_PROXY=${HTTPS_PROXY_URL}" "NO_PROXY=${NO_PROXY}"
 __HTTPS_DOCKER_PROXY__
+        else
+	    echo -e "\e[91mInvalid HTTPS Proxy URL supplied" > /dev/console
+        fi
     fi
 fi
--- a/manual/photon.xml.template
+++ b/manual/photon.xml.template
@@ -37,11 +37,11 @@
       <Category>Proxy Settings (optional)</Category>
         <Property ovf:key="guestinfo.http_proxy" ovf:type="string" ovf:userConfigurable="true">
             <Label>HTTP Proxy</Label>
-            <Description>Enter HTTP Proxy Server followed by the port and without typing "http://" before. Example: "proxy.provider.com:3128"</Description>
+            <Description>Enter HTTP Proxy URL followed by the port. Example: "http://proxy.provider.com:3128"</Description>
         </Property>
         <Property ovf:key="guestinfo.https_proxy" ovf:type="string" ovf:userConfigurable="true">
             <Label>HTTPS Proxy</Label>
-            <Description>Enter HTTPS Proxy Server followed by the port and without typing "https://" before. Example: "proxy.provider.com:3128"</Description>
+            <Description>Enter HTTPS Proxy URL followed by the port. Example: "https://proxy.provider.com:3128"</Description>
         </Property>
         <Property ovf:key="guestinfo.proxy_username" ovf:type="string" ovf:userConfigurable="true">
             <Label>Proxy Username (optional)</Label>


dsbibby:
Apologies - I've tried to sign my commits to this PR as described but my github email privacy settings have messed things up I think. I've fixed that now but I'm not sure how resolve commit 29128ea

pksrc:
@dsbibby one way is to rebase and amend the commit that is not signed. https://www.digitalocean.com/community/tutorials/how-to-rebase-and-update-a-pull-request

dsbibby:
@pksrc, thank you - I'm still getting to grips with git!

I think I've succeeded - dco-required label is removed at least an both my commits rebased in to one now.

files/setup-02-proxy.sh
```diff
@@ -23,10 +23,16 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
     fi
 
     if [ ! -z "${HTTP_PROXY}" ]; then
+        HTTP_PROXY_PREFIX="http"
+        a=(`printf '%s\n' "${HTTP_PROXY//\:\/\//$'\n'}"`)
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTP_PROXY_PREFIX=${a[0]}
+            HTTP_PROXY=${a[1]}
+        fi
```
lamw, line 31: In speaking with the team about this PR, I think it makes sense to NOT to assume whether it's HTTP(s) and let customers decide. This means this logic isn't needed and I'm suggesting that we change the original behavior from specifying <proxy>:<port> to the full URL of your proxy server whether that's HTTP or HTTPS.

files/setup-02-proxy.sh
```diff
@@ -36,15 +42,21 @@ __HTTP_DOCKER_PROXY__
     fi
 
     if [ ! -z "${HTTPS_PROXY}" ]; then
+        HTTPS_PROXY_PREFIX="https"
+        a=(`printf '%s\n' "${HTTPS_PROXY//\:\/\//$'\n'}"`)
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTPS_PROXY_PREFIX=${a[0]}
+            HTTPS_PROXY=${a[1]}
+        fi
```
lamw, line 50: Same as previous comment. Not needed

manual/photon.xml.template
```diff
@@ -37,11 +37,11 @@
       <Category>Proxy Settings (optional)</Category>
         <Property ovf:key="guestinfo.http_proxy" ovf:type="string" ovf:userConfigurable="true">
             <Label>HTTP Proxy</Label>
-            <Description>Enter HTTP Proxy Server followed by the port and without typing "http://" before. Example: "proxy.provider.com:3128"</Description>
+            <Description>Enter HTTP Proxy Server followed by the port and optionally typing "http://" before. Example: "proxy.provider.com:3128"</Description>
```
lamw, line 40: Please replace with:

> HTTP Proxy URL followed by the port. Example: "http://proxy.provider.com:3128"

manual/photon.xml.template
```diff
@@ -37,11 +37,11 @@
...
         </Property>
         <Property ovf:key="guestinfo.https_proxy" ovf:type="string" ovf:userConfigurable="true">
             <Label>HTTPS Proxy</Label>
-            <Description>Enter HTTPS Proxy Server followed by the port and without typing "https://" before. Example: "proxy.provider.com:3128"</Description>
+            <Description>Enter HTTPS Proxy Server followed by the port and optionally typing "http(s)://" before. Example: "https://proxy.provider.com:3128"</Description>
```
lamw, line 44: Please replace with

> HTTPS Proxy URL followed by the port. Example: "http://proxy.provider.com:3128"

files/setup-02-proxy.sh
dsbibby: Makes sense, however I think we still need to parse the proxy URL to remove the prefix so that credentials can be inserted if required?We could remove the conditional, but it might be better to add an else clause to log a failure (i.e. the user has entered a malformed URL)? I'll commit a change along these lines for you to review.

lamw: Not sure I follow the credentials comment? `HTTP_PROXY` and `HTTPS_PROXY` is only expected to have something like "http://proxy:3128" or "https://proxy:3128". Its not expected to contain any credentials. 

Proxy credentials are captured in `PROXY_USERNAME` and `PROXY_PASSWORD` and if those variables are detected, then you'd concat the final URI string which is then passed into the respective files

dsbibby: Apologies, perhaps I could have made that clearer! 

In the case of credentials being supplied, if we don't extract the users chosen protocol (`http` or `https`) from the passed `HTTP_PROXY` and `HTTPS_PROXY` variables, then we don't know what protocol to use when doing the concat of the final URI. 
Line 33 in my commit shows the issue: `HTTP_PROXY_URL="${HTTP_PROXY_PREFIX}://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY}"`

Unless I'm mistaken, we need to split the `HTTP_PROXY` URL on the `://` delimiter so that we can assemble the whole URI. Admittedly we only need to do this _if_ credentials are passed. However as suggested above we could take the opportunity here to present an error in the case of a malformed URL being given by the user.

lamw: Ah, I see what you mean now. Yes, you're correct and we can certainly throw an error if http(s) isn't detected. Your suggestion makes perfect sense and if you wan to update the implementing, we should be good!

lamw:
@dsbibby You can ensure your commits are always signed by using `git commit -s` when you commit a new change :)

dsbibby:
@lamw - good tip, thank you! 🙂

dsbibby:
@lamw - I think this is good for review again now. I'm not sure if the error handling is the best way approach? It certainly works, but is easily missed.

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
-            HTTP_PROXY_URL="http://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY}"
-        else
-            HTTP_PROXY_URL="http://${HTTP_PROXY}"
-        fi
-        echo "HTTP_PROXY=\"${HTTP_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/http-proxy.conf << __HTTP_DOCKER_PROXY__
+        a=(`printf '%s\n' "${HTTP_PROXY//\:\/\//$'\n'}"`)
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTP_PROXY_PREFIX=${a[0]}
+            HTTP_PROXY=${a[1]}
```
lamw, line 29: For ease of debugging, would you mind creating a brand new variable like `HTTP_PROXY_SERVER_PORT` rather than reusing `HTTP_PROXY`? I had to take a closer look at the code to see what was happening and was confused on why HTTP_PROXY was being redefined.

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
+            if [ $YES_CREDS -eq 1 ]; then
+                HTTP_PROXY_URL="${HTTP_PROXY_PREFIX}://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY}"
```
lamw, line 31: You'll need to replace `HTTP_PROXY` with the new variable mentioned above

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
+            else
+                HTTP_PROXY_URL="${HTTP_PROXY_PREFIX}://${HTTP_PROXY}"
```
lamw, line 33: You'll need to replace `HTTP_PROXY` with the new variable mentioned above

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
-            HTTPS_PROXY_URL="https://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTPS_PROXY}"
-        else
-            HTTPS_PROXY_URL="https://${HTTPS_PROXY}"
-        fi
-        echo "HTTPS_PROXY=\"${HTTPS_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/https-proxy.conf << __HTTPS_DOCKER_PROXY__
+        a=(`printf '%s\n' "${HTTPS_PROXY//\:\/\//$'\n'}"`)
+        if [ ${#a[*]} -eq 2 ]; then
+            HTTPS_PROXY_PREFIX=${a[0]}
+            HTTPS_PROXY=${a[1]}
```
lamw, line 49: For ease of debugging, would you mind creating a brand new variable like `HTTPS_PROXY_SERVER_PORT` rather than reusing `HTTPS_PROXY`? I had to take a closer look at the code to see what was happening and was confused on why HTTPS_PROXY was being redefined.

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
+            if [ $YES_CREDS -eq 1 ]; then
+                HTTPS_PROXY_URL="${HTTPS_PROXY_PREFIX}://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTPS_PROXY}"
```
lamw, line 51: You'll need to replace HTTPS_PROXY with the new variable mentioned above

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
+            else
+                HTTPS_PROXY_URL="${HTTPS_PROXY_PREFIX}://${HTTPS_PROXY}"
```
lamw, line 53: You'll need to replace HTTPS_PROXY with the new variable mentioned above

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
 
     if [ ! -z "${HTTP_PROXY}" ]; then
-        if [ $YES_CREDS -eq 1 ]; then
-            HTTP_PROXY_URL="http://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTP_PROXY}"
-        else
-            HTTP_PROXY_URL="http://${HTTP_PROXY}"
-        fi
-        echo "HTTP_PROXY=\"${HTTP_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/http-proxy.conf << __HTTP_DOCKER_PROXY__
+        a=(`printf '%s\n' "${HTTP_PROXY//\:\/\//$'\n'}"`)
```
lamw, line 26: For readability, can you replace tick `` marks with `$()` 

e.g.
```
a=($(printf '%s\n' "${HTTP_PROXY//\:\/\//$'\n'}"))
```

Ref: https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

files/setup-02-proxy.sh
```diff
@@ -23,28 +23,42 @@ if [ -n "${HTTP_PROXY}" ] || [ -n "${HTTPS_PROXY}" ]; then
...
 
     if [ ! -z "${HTTPS_PROXY}" ]; then
-        if [ $YES_CREDS -eq 1 ]; then
-            HTTPS_PROXY_URL="https://${PROXY_USERNAME}:${PROXY_PASSWORD}@${HTTPS_PROXY}"
-        else
-            HTTPS_PROXY_URL="https://${HTTPS_PROXY}"
-        fi
-        echo "HTTPS_PROXY=\"${HTTPS_PROXY_URL}\"" >> ${PROXY_CONF}
-        cat > ${DOCKER_PROXY}/https-proxy.conf << __HTTPS_DOCKER_PROXY__
+        a=(`printf '%s\n' "${HTTPS_PROXY//\:\/\//$'\n'}"`)
```
lamw, line 46: For readability, can you replace tick `` marks with `$()` 

e.g.
```
a=($(printf '%s\n' "${HTTPS_PROXY//\:\/\//$'\n'}"))
```

Ref: https://stackoverflow.com/questions/9449778/what-is-the-benefit-of-using-instead-of-backticks-in-shell-scripts

lamw:
@dsbibby Once you've addressed the latest changes, feel free to squash all commits into one and we should be good to merge.

dsbibby:
@lamw - all sensible suggestions - I renamed the `HTTP_PROXY_PREFIX` variables to `HTTP_PROXY_PROTOCOL` also as it's more descriptive I think.

lamw:
@dsbibby The updates looks good, thanks! 

One final question, did you build the appliance to make sure all changes worked as expected? We normally use this a final confirmation as part of the PR.

dsbibby:
@lamw I hadn't (I'd just tested in place in an exisitng appliance) but building properly is the correct thing to do, and sounded fun, so I have now!

All seems to work as expected. I've deployed in to an environment that needs proxies and one that doesn't and both work as expected. Also, passing a malformed URI as a proxy field logs an error during first boot of the appliance.

lamw:
@dsbibby We just realized this PR was against `master` rather than `development` which is the normal process. We'll need to revert the merge but wanted to ensure you still have your branch to resubmit. Please ACK and we'll handle the revert and you can submit a new PR against the development branch

dsbibby:
@lamw Yes, I still have the branch. Go ahead and revert and I'll submit a new PR when you're ready.

lamw:
@dsbibby Thanks. We've reverted, go ahead and submit a new PR against development branch and we'll get that merged up.

dsbibby:
@lamw all done. I've submitted a new PR #161

PR #146, opened 2020-05-15
Introduce Error struct in processor

martindekov:
Introducing Error struct in processor package next to the
interface, to be used by the Process method to extend the
errors with more context by providing the name of the
processor

Signed-off-by: Martin Dekov <mvdekov@gmail.com>

### DIFF
--- a/vmware-event-router/internal/processor/aws_event_bridge.go
+++ b/vmware-event-router/internal/processor/aws_event_bridge.go
@@ -3,6 +3,7 @@ package processor
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
 	"os"
 	"sync"
@@ -176,8 +177,9 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObjectReference, baseEvent []types.BaseEvent) error {
 	batchInput, err := awsEventBridge.createPutEventsInput(baseEvent)
 	if err != nil {
-		awsEventBridge.Printf("could not create PutEventsInput for event(s): %v", err)
-		return nil
+		errMsg := fmt.Errorf("could not create PutEventsInput for event(s): %v", err)
+		awsEventBridge.Printf(errMsg.Error())
+		return NewError(ProviderAWS, errMsg).Error()
 	}
 
 	// nothing to send
--- a/vmware-event-router/internal/processor/aws_event_bridge.go
+++ b/vmware-event-router/internal/processor/aws_event_bridge.go
@@ -178,7 +178,7 @@ func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObject
 	batchInput, err := awsEventBridge.createPutEventsInput(baseEvent)
 	if err != nil {
 		errMsg := fmt.Errorf("could not create PutEventsInput for event(s): %v", err)
-		awsEventBridge.Printf(errMsg.Error())
+		awsEventBridge.Println(errMsg)
 		return processorError(ProviderAWS, errMsg)
 	}
 
--- a/vmware-event-router/internal/processor/aws_event_bridge.go
+++ b/vmware-event-router/internal/processor/aws_event_bridge.go
@@ -179,7 +179,7 @@ func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObject
 	if err != nil {
 		errMsg := fmt.Errorf("could not create PutEventsInput for event(s): %v", err)
 		awsEventBridge.Printf(errMsg.Error())
-		return NewError(ProviderAWS, errMsg).Error()
+		return processorError(ProviderAWS, errMsg)
 	}
 
 	// nothing to send
--- a/vmware-event-router/internal/processor/processor.go
+++ b/vmware-event-router/internal/processor/processor.go
@@ -22,12 +22,11 @@ type Error struct {
 	err       error
 }
 
-// NewError constructs Error structure
-func NewError(processor string, err error) *Error {
+func processorError(processor string, err error) error {
 	return &Error{
 		processor: processor,
 		err:       err,
 	}
 }
 
-func (e *Error) Error() error { return fmt.Errorf("%s: %s", e.processor, e.err.Error()) }
+func (e *Error) Error() string { return fmt.Sprintf("%s: %s", e.processor, e.err.Error()) }
--- a/vmware-event-router/internal/processor/processor.go
+++ b/vmware-event-router/internal/processor/processor.go
@@ -1,6 +1,8 @@
 package processor
 
 import (
+	"fmt"
+
 	"github.com/vmware/govmomi/vim25/types"
 )
 
@@ -11,3 +13,21 @@ import (
 type Processor interface {
 	Process(types.ManagedObjectReference, []types.BaseEvent) error
 }
+
+// Error struct contains the generic error content used by the processors
+// it extends the simple error by providing context which processor gave
+// the error
+type Error struct {
+	processor string
+	err       error
+}
+
+// NewError constructs Error structure
+func NewError(processor string, err error) *Error {
+	return &Error{
+		processor: processor,
+		err:       err,
+	}
+}
+
+func (e *Error) Error() error { return fmt.Errorf("%s: %s", e.processor, e.err.Error()) }
--- a/vmware-event-router/internal/processor/processor_test.go
+++ b/vmware-event-router/internal/processor/processor_test.go
@@ -1,0 +1,36 @@
+package processor
+
+import (
+	"fmt"
+	"testing"
+)
+
+func Test_processor_error(t *testing.T) {
+	tests := []struct {
+		title       string
+		provider    string
+		errMessage  error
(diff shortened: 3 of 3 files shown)


vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -176,8 +177,9 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObjectReference, baseEvent []types.BaseEvent) error {
 	batchInput, err := awsEventBridge.createPutEventsInput(baseEvent)
 	if err != nil {
-		awsEventBridge.Printf("could not create PutEventsInput for event(s): %v", err)
-		return nil
+		errMsg := fmt.Errorf("could not create PutEventsInput for event(s): %v", err)
+		awsEventBridge.Printf(errMsg.Error())
+		return NewError(ProviderAWS, errMsg).Error()
```
embano1, line 182: That should be sufficient:

```suggestion
		return NewError(ProviderAWS, errMsg)
```

embano1, line 182: Can you please provide an example of a formatted error string response in the comment of this PR?

vmware-event-router/internal/processor/processor.go
```diff
@@ -11,3 +13,21 @@ import (
...
+
+// NewError constructs Error structure
+func NewError(processor string, err error) *Error {
+	return &Error{
+		processor: processor,
+		err:       err,
+	}
+}
+
+func (e *Error) Error() error { return fmt.Errorf("%s: %s", e.processor, e.err.Error()) }
```
embano1, line 33: We can simply satisfy the error interface:

```go
type error interface {
    Error() string
}
```

vmware-event-router/internal/processor/processor.go
```diff
@@ -11,3 +13,21 @@ import (
...
+// Error struct contains the generic error content used by the processors
+// it extends the simple error by providing context which processor gave
+// the error
+type Error struct {
+	processor string
+	err       error
+}
+
+// NewError constructs Error structure
+func NewError(processor string, err error) *Error {
```
embano1, line 26: Do we need to export this one? 
Also: what do you think of renaming this to:

```go
func ProcessorError(processor string, err error) error {}
```

This would then be used via `return ProcessorError(ProviderAWS, errMsg)`

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Will remove 👍

martindekov: Will do 👍

vmware-event-router/internal/processor/processor.go
martindekov: You are right, will replace 👍

vmware-event-router/internal/processor/processor.go
martindekov: Well it is in the same package, if we are going to use the error in the processors only there is no need to export it. It also looks prettier that way 😄 will lowercase it so it is not exported outside the package. Otherwise will change as proposed 👍

martindekov:
On a side note @embano1 I though there would be more places where we should clearly exit with an error and stop the processing I found only one and put the error exit there. Otherwise we process multiple events which we should re-try instead of failing the whole cycle when. I suppose returning and error would mean that we just cancel the whole batch of events. We should probably think how to signal events to be retried once we have at least once semantics.

embano1:
> there would be more places where we should clearly exit with an error and stop the processing

When you say `exit` you mean returning or `exit != 0`?

> I found only one and put the error exit there. 

What would be other places where you would like to see this?

> Otherwise we process multiple events which we should re-try instead of failing the whole cycle when. I suppose returning and error would mean that we just cancel the whole batch of events. We should probably think how to signal events to be retried once we have at least once semantics.

Yes, this is a problem as of today we need to decide how to handle with the at most once semantics we have today (e.g. fail batch, batch event, all batches). What do you think would be a good UX?

vmware-event-router/internal/processor/processor.go
martindekov: Returns string (error interface) ✔️

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Added unit testing which show the expected output ✔️

martindekov: No `.Error()` now ✔️

vmware-event-router/internal/processor/processor.go
martindekov: Internal for the package ✔️

martindekov:
> When you say exit you mean returning or exit != 0?


> Yes, this is a problem as of today we need to decide how to handle with the at most once semantics we have today (e.g. fail batch, batch event, all batches). What do you think would be a good UX?

Return error when the processor is unreachable, not skip the event which failed to be processed (malformed event for example). In general in my mind having failed to process and event and the processor being down in a way (unreachable?) are two separate things:
* At least once: Change interface a little to return struct of failed events `FailedEvents` (possibly?) to get the events which should be retried? Not sure yet but this is in my mind:
```
type Processor interface {
	Process(types.ManagedObjectReference, []types.BaseEvent) (FailedEvents, error)
}
```
* processorError: Return to notify that the processor itself has some problem (OF gateway might be unreachable? AWS EventBridge service might be down?) (problem is not in event, but processor itself) Note we might still want to have `processorError` + `FailedEvents` returned in this situation so we can retry the events even though the processors are unreachable in a way.

Hope this sounds reasonable ^

> What would be other places where you would like to see this?

I was just left with the impression that we return error from the `Process` more often with the Processor being the main fault, not the event, I guess my impression was wrong.

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -176,8 +177,9 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
 func (awsEventBridge *awsEventBridgeProcessor) Process(moref types.ManagedObjectReference, baseEvent []types.BaseEvent) error {
 	batchInput, err := awsEventBridge.createPutEventsInput(baseEvent)
 	if err != nil {
-		awsEventBridge.Printf("could not create PutEventsInput for event(s): %v", err)
-		return nil
+		errMsg := fmt.Errorf("could not create PutEventsInput for event(s): %v", err)
+		awsEventBridge.Printf(errMsg.Error())
```
embano1, line 181:
```suggestion
		awsEventBridge.Println(errMsg)
```

embano1: ^ is that suggestion ok?

vmware-event-router/internal/processor/processor_test.go
```diff
@@ -0,0 +1,36 @@
+package processor
```
embano1, line 1:
```suggestion
// +build unit

package processor
```

vmware-event-router/internal/processor/aws_event_bridge.go
embano1: Do we need to change lines in OpenFaaS proc?

embano1:
> Return error when the processor is unreachable, not skip the event which failed to be processed (malformed event for example). In general in my mind having failed to process and event and the processor being down in a way (unreachable?) are two separate things

You are spot on and I fully agree that these error cases have to be handled separately, e.g. temporary/fatal errors. Here's couple of thoughts on this:

- most fatal cases are caught during initialization of the processor - if after initialization a critical error occurs in the processor, often we can't do much because the library itself might panic (instead of indicating different error cases)
- even such critical errors I would consider "non-fatal" from the point of view of the event router because they could still be temporary, e.g. backoff logic retrying when gateway not available and buffer events (especially when we want to support multiple providers/processors later)
- the error returned by the processor implementation might not be distinguishable between error due to gateway not available or error due to function
- AWS processor does not really continuously run as a backend process in the router (only periodic syncs), i.e. all RPCs are synchronous when an event fires and matches the subscription pattern
- OpenFaaS `Controller.Invoke` does not return error immediately, rather through `ResponseSubscriber interface` so we can't capture the error in `Process(...)`

My suggestion would be to treat any processor error after successful initialization of a particular processor as temporary with logging/retrying/DLQ logic specific to the processor. Since error handling is very specific to the actual processor implementation, e.g. see difference between AWS and OpenFaaS, the question is whether error handling should always be the concern of the processor or the event router. If the former I can't really think of many errors that would need to be handled via the router itself.

vmware-event-router/internal/processor/aws_event_bridge.go
martindekov: Changed to `awsEventBridge.Println(errMsg)` ✔️  The openfaas error comes in the struct not through Invoke method, so we should think how exactly those errors should be added to the OF processors

vmware-event-router/internal/processor/processor_test.go
martindekov: Added ✔️

PR #152, opened 2020-05-19
install gems only for the user, and set a vendor path for bundle use

tompscanlan:
Using a mac, I needed to do this to test a change in the docs.

* `gem ... --user-install` to avoid using sudo
* use a vendor dir that is outside of the docs tree

Signed-off-by: Tom Scanlan <tscanlan@vmware.com>

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,7 +24,7 @@ The website is hosted using [Github Pages](https://help.github.com/en/github/wor
 └── Gemfile                   > Plugins required for the website to be built by Jekyll
 ```
 
-In order for Jekyll to process the MD files and render them as html, you'll need to add the below to the beginning of the each MD file. 
+In order for Jekyll to process the MD files and render them as html, you'll need to add the below to the beginning of the each MD file.
 
 ```yaml
 ---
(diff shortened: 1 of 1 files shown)


tompscanlan:
@pksrc that would be fine.  let me update and see if it makes more sense.

docs/README.md
```diff
@@ -62,6 +62,8 @@ rbenv install 2.6.3
 gem install bundler
 ```
 
+*Note: if you hit a permissions error for the `gem install bundler` follow advice from the [bundler docs](https://bundler.io/doc/troubleshooting.html#permission-denied-when-installing-bundler)*
```
pksrc, line 65: This looks good. However, I noticed this link does not talk about the `vendor/cache/` which you had mentioned w/ the first change. Let me know if that was required for you to build the website along with the user install?

tompscanlan: Turns out that I didn't need to change the vendor cache location.  default seems to be fine.

pksrc:
@tompscanlan My apologies but I forgot to have you squash the commits before merging. Since the changes in the PR are related and can be referenced with a single commit, I need to revert the merge. Please ACK if you can resubmit the PR with the commits squashed.

Issue #160, opened 2020-05-25
Processor Interface should not be coupled to vCenter


embano1:
Change

```go
type Processor interface {
	Process(types.ManagedObjectReference, []types.BaseEvent) error
}
```

To use `CloudEvents` spec:

```go
type Processor interface {
	Process([]cloudevent.Event) error
}
```

Issue #120, opened 2020-05-06
Better error handling in the processors


martindekov:
Currently we handle errors coming only from Events coming from vcenter and not the callback functions themselves:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/9a9ef34c867dd8e814fd1dafb80a27d0c1d82e85/vmware-event-router/internal/stream/vcenter.go#L88

This can be seen in both processors:
* OpenFaaS:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/9a9ef34c867dd8e814fd1dafb80a27d0c1d82e85/vmware-event-router/internal/processor/openfaas.go#L111
* AWS:
https://github.com/vmware-samples/vcenter-event-broker-appliance/blob/9a9ef34c867dd8e814fd1dafb80a27d0c1d82e85/vmware-event-router/internal/processor/aws_event_bridge.go#L176

We should create custom errors by Implementing the Error method:
```go
type ProcessorErr struct {
    processor string
    content string
}
```

Implementing the Error interface:
```go
func (e *ProcessorErr) Error() string { return ....}
```
and the same for OpenFaaS and every processor afterwards. Instead of returning `nil` return the error in the context of: `ProcessorErr.Error("event_bridge","error content")`.

This will enable better integration testing of the event router.

embano1:
Thank you @martindekov ! Can we abstract the error so we don't have a type for every processor? E.g. create a `processor.Error` type in `processor.go` and use across all processors with a constructor `Error.New(processor string, err error)`?

martindekov:
Yeah Michael no problems followed through in the explanation :+1:

embano1:
Closed: #146

PR #166, opened 2020-05-29
Decouple from types.BaseEvent

embano1:
This marks the first commit in a series to decouple the VMware Event
Router processing logic from vCenter. Specifically, this commit changes
the processor interface to using fully leveraging CloudEvents v1 Spec
(SDKv2 package). Before invoking a processor stream provider events,
e.g. vCenter, are converted to CloudEvents v1 spec and validated.

The (vCenter) stream provider now invokes each processor via a single
CloudEvent (instead of []types.BaseEvent). Batching, e.g. as was in
EventBridge, could be re-implemented at a later stage (see note on state
management below). Both, stream providers and stream processors have
better error handling (depending on the implementation though).

Note that there is still coupling between (vCenter) stream provider
logic to the processor which need to be addressed in separate PR(s) via
state management.

Tested with OpenFaaS and AWS EventBridge stream processor.

Closes: #160 
Supersedes: #120 


vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -171,102 +178,70 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
+		awsEventBridge.Printf("processing event (ID %s): %v", ce.ID(), ce)
 	}
 
-	for idx, input := range batchInput {
-		resp, err := awsEventBridge.PutEvents(&input)
-		if err != nil {
-			awsEventBridge.Printf("could not send event(s) for batch %d: %v", idx, err)
-			continue
-		}
+	awsEventBridge.mu.RLock()
```
martindekov, line 187: Do we lock here since we process single cloudevent in Process now?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -171,102 +178,70 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-	// nothing to send
-	if len(batchInput) == 0 {
-		return nil
+// Process implements the stream processor interface
+func (awsEventBridge *awsEventBridgeProcessor) Process(ce cloudevents.Event) error {
+	if awsEventBridge.verbose {
+		awsEventBridge.Printf("processing event (ID %s): %v", ce.ID(), ce)
 	}
 
-	for idx, input := range batchInput {
```
martindekov, line 190: Does moving to cloudevents mean we no longer process every event as is(stacking it in the processor)?

martindekov: Relying on event bridge to handle this for us?

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -171,102 +178,70 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
-	defer awsEventBridge.mu.Unlock()
 
-	batch := []eventbridge.PutEventsInput{}
-
-	tmpInput := eventbridge.PutEventsInput{
-		Entries: []*eventbridge.PutEventsRequestEntry{},
+	jsonBytes, err := json.Marshal(ce)
+	if err != nil {
+		msg := fmt.Errorf("could not marshal event %v: %v", ce, err)
+		awsEventBridge.Println(msg)
```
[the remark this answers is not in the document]
embano1: Yeah, I also thought this would be too much since we bubble the error up. Left it in since it's a processor error and currently we don't have providers and processors decoupled. So logging this error at stream provider level (e.g. vCenter) seemed a bit misleading to me. What do you think?

vmware-event-router/internal/processor/aws_event_bridge.go
embano1: Yes we do (and need the lock if that's the question) since there's potential concurrent access to the map due to the background goroutine syncing the rules.

vmware-event-router/internal/processor/aws_event_bridge.go
embano1: Not sure I understand the question. Is it about batching (previous version) vs non-batching (this PR)?

martindekov: I am sorry my comment was ambiguous. But in general yeah, my question was about removing the batching. But I saw:
```
Batching, e.g. as was in
EventBridge, could be re-implemented at a later stage (see note on state
management below)
.```
So I guess this answers my question

File: vmware-event-router/internal/processor/aws_event_bridge.go (hunk 1 above)
Comment (martindekov) [reply vmware-event-router/internal/processor/aws_event_bridge.go:187]: Yeah, ok I was just curious as this lock is new 👍

Comment (martindekov) [reply vmware-event-router/internal/processor/aws_event_bridge.go:187]: Never mind the new, saw we use the lock in another method

File: vmware-event-router/internal/processor/aws_event_bridge.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (martindekov) [reply vmware-event-router/internal/processor/aws_event_bridge.go:200]: Yeah in that context it seems reasonable to leave this as is for now I just saw in `vcenter_fake`:
```
				if err != nil {
					f.Logger.Printf("could not process event %v: %v", ce, err)
				}
```
And though we would just print it twice if there is an error. Your comment is valid so let's leave this as is

File: vmware-event-router/internal/processor/aws_event_bridge.go [hunk 4]
```diff
@@ -171,102 +178,70 @@ func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, sour
...
+	input := eventbridge.PutEventsInput{
+		Entries: []*eventbridge.PutEventsRequestEntry{&entry},
 	}
-	if len(tmpInput.Entries) > 0 {
-		batch = append(batch, tmpInput)
+	awsEventBridge.Printf("sending event %s", ce.ID())
+	resp, err := awsEventBridge.PutEvents(&input)
+	if err != nil {
+		msg := fmt.Errorf("could not send event %v: %v", ce, err)
+		awsEventBridge.Println(msg)
```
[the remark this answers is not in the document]
martindekov: Discussed above

vmware-event-router/cmd/main.go
```diff
@@ -90,7 +90,7 @@ func main() {
 			switch cfg.Provider {
 			case stream.ProviderVSphere:
 				logger.Printf("connecting to vCenter %s", cfg.Address)
-				streamer, err = stream.NewVCenterStream(ctx, cfg, metricsServer)
+				streamer, err = stream.NewVCenterStream(ctx, cfg, metricsServer, stream.WithVCenterVerbose(verbose))
```
codegold79, line 93: All right--making use of function options!

vmware-event-router/internal/processor/aws_event_bridge.go
```diff
@@ -53,13 +55,18 @@ type eventPattern struct {
 
 // NewAWSEventBridgeProcessor returns an AWS EventBridge processor for the given
 // stream source.
-func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, source string, verbose bool, ms *metrics.Server) (Processor, error) {
+func NewAWSEventBridgeProcessor(ctx context.Context, cfg connection.Config, ms metrics.Receiver, opts ...AWSOption) (Processor, error) {
```
codegold79, line 58: This is one massive construction function. Not only does it have multiple switch statements and for loops, it also runs two goroutines. I would recommend breaking this function up, maybe putting the section finding the ARN rules in another function. But, I can see the function works, so I'm also fine with keeping it. Even if it's a long function, it's not too difficult to follow what's going on.

embano1: Fair point, even though I've seen larger constructors in the wild :) I definitely think breaking out the options handling and pattern match logic into helpers makes sense, especially since the `syncRules` has a lot of repetitive code here.

codegold79: Agreed that the change would better suited to be in a separate PR. I have raised a new issue to have the matter addressed: https://github.com/vmware-samples/vcenter-event-broker-appliance/issues/168

PR #181, opened 2020-06-18
Updated Fling URL from vcenter to vmware

kremerpatrick:
Link change to reference the new flings URL  vmware-event-broker-appliance instead of vcenter-event-broker-appliance. The redirect works currently but we might as well reference the correct link.

Tested by clicking the link on local webserver.


pksrc:
LGTM

PR #184, opened 2020-07-10
Example Pre-Filter function

dsbibby:
This function allows you to use regex filters to match the inventory paths of any managed object references received from vCenter. If everything matches the filters you define, the function will trigger a definable chained function. So it can be used in front of any existing function to limit the scope of when that function is run - e.g. If you only want to tag VMs within a specific folder or resource pool when they're powered on.

- The function uses the python3-flask language template to allow a persistent connection to vCenter.
- vCenter connection information is contained in the `vc-secrets.toml` file
- All other config (filters, chained function name) is contained in the `stack.yml` file as environment variables to allow multiple functions to be defined.
- The `filter_...` environment variables are regex filters that are named after the event properties of vCenter managed objects in the event payload (e.g. `filter_vm`, `filter_datacenter`, `filter_computeresource`, etc.). The regex is tested against the full infrastructure path of the related managed object, with a default if not set of `. *`. 
- The `call_function` environment variable is the name of an existing deployed function. 

Would welcome any feedback on this, many thanks! 

Signed-off-by: David Bibby <davidsbibby@gmail.com>


embano1:
Thank you David for your contribution and sorry for my late reply. I will try to review this PR ASAP.

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+    vcinfo = service_instance.RetrieveServiceContent().about
+    debug(f'Connected to {vcinfo.fullName} ({vcinfo.instanceUuid})')
+
+    # Load the Events that function gets from vCenter through the Event Router
+    debug(f'{bgc.HEADER}Reading Cloud Event: {bgc.ENDC}')
+    debug(f'{bgc.OKBLUE}Event > {bgc.ENDC}{req}')
+    try:
+        cevent = json.loads(req)
+    except json.JSONDecodeError as err:
+        return 'Invalid JSON > JSONDecodeError: {0}'.format(err), 400
```
embano1, line 174:
```suggestion
        return 'Invalid JSON > JSONDecodeError: {0}'.format(err), 500
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+
+    debug(f'{bgc.HEADER}Validating Input data: {bgc.ENDC}')
+    debug(f'{bgc.OKBLUE}Event > {bgc.ENDC}{json.dumps(cevent, indent=4, sort_keys=True)}')
+    try:
+        # CloudEvent - simple validation
+        event = cevent['data']
+    except KeyError as err:
+        traceback.print_exc(limit=1, file=sys.stderr)  # providing traceback since it helps debug the exact key that failed
+        return 'Invalid JSON, required key not found > KeyError: {0}'.format(err), 400
```
embano1, line 183:
```suggestion
        return 'Invalid JSON, required key not found > KeyError: {0}'.format(err), 500
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+    except AttributeError as err:
+        traceback.print_exc(limit=1, file=sys.stderr)  # providing traceback since it helps debug the exact key that failed
+        return 'Invalid JSON, data not iterable > AttributeError: {0}'.format(err), 400
```
embano1, line 186:
```suggestion
        return 'Invalid JSON, data not iterable > AttributeError: {0}'.format(err), 500
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+
+    debug(f'{bgc.HEADER}All tests passed! Testing all object filters:{bgc.ENDC}')
```
embano1, line 188:
```suggestion
    debug(f'{bgc.HEADER}Validation passed! Applying object filters:{bgc.ENDC}')
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+def handle(req):
+    """
+    Handle a request to the function
+    Args:
+        req (str): request body
+    """
+
+    # Initialise a connection to vCenter
+    res = init()
+    if isinstance(res, tuple):  # Error state
```
embano1, line 163:
```suggestion
    if isinstance(res, tuple):  # Tuple is returned if an error occurred
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+    while obj != service_instance.content.rootFolder:
+        path = f'/{obj.name}{path}'
+        obj = obj.parent
+    return path
+
+
+def filterVcObject(obj, filter):
+    """
+    Takes a VC managed object and tests it against a filter
+    If it matches, a vRO input parameter object is returned
```
embano1, line 136: "vRO input parameter"?

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+                    # Get a vc managed object
+                    mo = getManagedObject(item)
+                    # Test the filter
+                    res = filterVcObject(mo, objFilter)
+                    if isinstance(res, tuple):  # Tuple is returned if the object didn't match the filter
+                        return res
+
+    func = os.getenv("call_function")
+    if func:
+        debug(f'{bgc.HEADER}All filters passed. Calling chained function {func}{bgc.ENDC}')
```
embano1, line 206:
```suggestion
        debug(f'{bgc.HEADER}Filter matched. Calling chained function {func}{bgc.ENDC}')
```

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+
+        try:
+            res = requests.post(f'http://gateway.openfaas:8080/function/{func}', req)
+            return res.text, res.status_code
+        except Exception as err:
+            traceback.print_exc(limit=1, file=sys.stderr)  # providing traceback since it helps debug the exact key that failed
+            sys.stderr.flush()
+            return 'Unexpected error occurred > Exception: {0}'.format(err), 500
+
+    # debug(f'Successfully executed vRO workflow: {vro_res["name"]}')
```
embano1, line 216: Can we delete this? (all vRO refs)

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -0,0 +1,260 @@
...
+    return "No function to call", 400
```
embano1, line 217:
```suggestion
    return "call_function must be specified", 400
```

examples/python/pre-filter/pre-filter/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pyvim
```
embano1, line 1: Do you want to pin the versions?

examples/python/pre-filter/stack.yml
```diff
@@ -0,0 +1,20 @@
+version: 1.0
+provider:
+  name: openfaas
+  gateway: https://veba.lab.local
```
embano1, line 4:
```suggestion
  gateway: https://VEBA_FQDN_OR_IP
```

examples/python/pre-filter/stack.yml
```diff
@@ -0,0 +1,20 @@
...
+functions:
+  pre-filter:
+    lang: python3-flask
+    handler: ./pre-filter
+    image: pilue/pre-filter:latest
```
embano1, line 9: Note: Once this PR is approved we need to change this to the VMware repo and a different container name.

dsbibby:
Hi @embano1, many thanks for the review - all easy stuff to resolve.

Apologies for the confusing vRO references - this started as a vRO function but it became obvious to me that it would be better and more useful to split it in to two functions. I'll tidy those up, and write up a readme - I was holding off that until I was sure you were happy with this as an example.

I'll push an update as soon as I can.

examples/python/pre-filter/stack.yml
dsbibby: Understood - let me know if there's anything I need to do here.

examples/python/pre-filter/pre-filter/handler.py
dsbibby: I don't know how/if this impacts on OpenFaaS handling of error states, but in this case a `JSONDecodeError` indicates a malformed request by the client rather than an internal error doesn't it?

Additionally, other python function examples return 400 for issues parsing the JSON request data.

examples/python/pre-filter/pre-filter/handler.py
dsbibby: Same as above... is a 400 code not correct here?

examples/python/pre-filter/pre-filter/handler.py
dsbibby: Same as above... is a 400 code not correct here?

examples/python/pre-filter/pre-filter/handler.py
embano1: Sorry for causing confusion with the existing examples and how they treat the JSON (event) payload. Looks like I need to clean them up as well :)

Since the event router create the JSON (event) payload and I was considering the function(s) written for this project as one unit (router + function) I was suggesting 500 to indicate something is wrong with the internal logic of the router. A 400 could trip off users, looking for different reasons where the actual reason in 99% of the cases must be the router. 

If we treat the functions as "standalone", then 400 makes sense (as in the older examples) to indicate an error with the supplied payload (even though that will likely come from the router itself).

examples/python/pre-filter/stack.yml
embano1: If you want, please take a look at the existing Python functions and suggest a name for the image.

dsbibby:
Hi @embano1, I have made all the suggested changes, and fixed a few issues along the way. Please could you review the changes, particularly the `README.md` for clarity? Many thanks!

examples/python/pre-filter/stack.yml
```diff
@@ -0,0 +1,19 @@
...
+  pre-filter:
+    lang: python3-flask
+    handler: ./pre-filter
+    image: vmware/veba-python-pre-filter:latest
+    environment:
+      write_debug: true
+      read_debug: true
+      insecure_ssl: true
+      call_function: veba-echo
+      filter_vm: .*
```
embano1, line 15: Does your script also work when using `'.*'` for the `filter_vm` spec? Reason I ask is because someone might accidentally set the filter to something like `*` which is invalid yaml (https://cloudslang-docs.readthedocs.io/en/v1.0/yaml_overview.html#scalars-strings-numbers-values).

docs/site/examples.md
```diff
@@ -130,6 +130,15 @@ examples:
     links:
     - language: powershell
       url: "/tree/master/examples/powershell/vmware-cloud-ngw-teams"
+
+  - title: vCenter Managed Object Pre-Filter
```
embano1, line 134: @pksrc can you please double-check?

examples/python/pre-filter/stack.yml
dsbibby: Yes, it does. I removed the quotes for the example pattern as I figured they could cause confusion - particularly if a pattern required quotes for some reason. That said `*` is also an invalid regex. However `[a-z]` is a valid regex but not valid yaml. I've just tried both - in the case of invalid yaml `faas-cli` won't let you deploy the function. However I found I wasn't catching invalid regex patterns in the function so I'll add handling for that case to a new commit. For the example you've highlighted here, I could add quotes to show that's an option, or leave them out and add a bit to the `README.md` explaining this.

examples/python/pre-filter/pre-filter/handler.py
```diff
@@ -150,12 +150,17 @@ def filterVcObject(obj, filter):
...
-        else:
-            debug(f'{bgc.OKBLUE}Match > {bgc.ENDC}Filter matched {moType} path')
-            return True
+        try:
+            if not re.search(filter, objPath):
+                debug(f'{bgc.WARNING}Filter "{filter}" does not match {moType} path "{objPath}". Exiting{bgc.ENDC}')
+                return f'Filter "{filter}" does not match {moType} path "{objPath}"', 200
+            else:
+                debug(f'{bgc.OKBLUE}Match > {bgc.ENDC}Filter matched {moType} path')
+                return True
```
embano1, line 159: Do you really want to return `True` here which seems to be used to return from the handler at a later point?

dsbibby: I don't think it matters really - I only test if the result is a tuple later in line 216, any other result is ignored. That said this return is superfluous so I'll delete it!

I'll commit that one change a squash shortly. Many thanks!

dsbibby:
Commits squashed and outstanding changes made. Also fixed a few typos in the `README.md`.

PR #326, opened 2021-04-02
Knative echo service example written in Go

codegold79:
Signed-off-by: F. Gold <fgold@vmware.com>

## Summary

Provide an example of a Knative echo service written in Go.

Working docker image, `fgold/kn-go-echo:0.1` has been built for testing purposes. After merge, people should use `veba/kn-go-echo:0.1`.

## Pull Request Checklist
🚨 Please review the [guidelines for contributing](https://vmweventbroker.io/community) to this repository.

- [x] Please ensure that you are making a pull request against the **Development** branch
- [x] Please use the `WIP` keyword in the title of your PR if you are not ready for review
- [ ] Please ensure that you have opened a Github Issue if you are resolving/fixing a problem
- [x] Please ensure that you have [signed](https://help.github.com/en/github/authenticating-to-github/signing-commits) all commits and that you have [squashed](https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec) all relevant commmits related to your change
- [x] Please make sure that you have tested your change locally by successfully [building and deploying the VMware Event Broker Appliance](https://vmweventbroker.io/kb/contribute-appliance) and/or [building and deploying VMware Event Router](https://vmweventbroker.io/kb/contribute-eventrouter)
- [x] Please include any relevant screenshots and/or output as part of your testing
- [x] Please include any documentation updates that is applicable for your changes

## Change Type

What types of changes does your code introduce to the VMware Event Broker Appliance?

Please check the type of change your PR introduces:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Documentation changes
- [x] Other (please describe): Knative example service.

## Testing Verification

* Successfully built echo function that is able to run locally as a docker image as well as on a Knative installed with [KonK](https://github.com/csantanapr/knative-kind).
* Successfully deployed service and trigger using `kubectl apply -f function.yaml -n vmware-functions` inside VEBA.
* Relevant screenshots and/or output as part of the testing
Screen Shot 2021-05-07 at 12 26 31 PM


examples/go/kn-echo/README.md
```diff
@@ -0,0 +1,54 @@
+# kn-echo
```
lamw, line 1: FYI - We're going to be simplifying the function examples by NOT using  the`kn` CLI but rather `kubectl` to deploy Knative Service/Trigger. The mains reason is that kn does not have support for secrets and most functions today will  contain some form of sensitive/secret entries.
 
You can take a look at the following 
https://github.com/lamw/vcenter-event-broker-appliance/tree/knative-docs/examples/knative/kn-echo to get an idea of what the new structure layout should follow

codegold79: Cool, thanks @lamw--will take a look.

lamw: @codegold79 Sorry, it looks like during the v0.6 release, Michael and I agreed that the using `pack` to build images may have some short comings and decided to hold back his Python function example to be more "Docker" friendly. For reference, here's what we did for PowerShell https://github.com/vmware-samples/vcenter-event-broker-appliance/tree/development/examples/knative/powershell/kn-ps-slack

@embano1 Did you want to have a chat w/Frankie to see what approach we want to go with for the non-PowerShell examples? We'll need this for other language contributions but golang would be a good one to baseline and then have one similar for Python

codegold79: I can ditch `pack` and provide a Dockerfile. I can have a discussion with @embano1 next week regarding what base images we should use: one for building the app and one for running it.

codegold79:
I've updated the kn-go-echo example to use go cloud-sdk per @embano1's recommendation.

TODO: remove `pack` and include a Dockerfile to run the example echo service.

embano1:
@codegold79 I was wondering if we should use [`ko`](https://github.com/google/ko) to ease the building and deployment of Go related functions? This would remove the need to write and maintain Dockerfiles.

Not sure what @lamw thinks about it. In the (only) Python example we currently have we use `pack` to abstract these details.

examples/knative/go/kn-go-echo/server.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"fmt"
+	"log"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+)
+
+func main() {
+	ctx := context.Background()
+	protocol, err := cloudevents.NewHTTP()
```
embano1, line 13: You could simplify this with `NewClientHTTP` (godoc: https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.4.1/client#NewHTTP)

examples/knative/go/kn-go-echo/server.go
```diff
@@ -0,0 +1,29 @@
...
+	client, err := cloudevents.NewClient(protocol)
+	if err != nil {
+		log.Fatalf("creating client, %v", err)
+	}
+
+	log.Printf("listening on :8080\n")
+	log.Fatalf("start receiver: %s", client.StartReceiver(ctx, receive))
+}
+
+func receive(ctx context.Context, event cloudevents.Event) {
```
embano1, line 27: Shall we take this to the next level and also return the received event in case this example is used as part of a pipeline?

examples/knative/go/kn-go-echo/server.go
```diff
@@ -0,0 +1,29 @@
...
+	fmt.Printf("***cloud event***\n%s", event)
```
embano1, line 28: Will this work nicely with different payload encoding schemes, e.g. XML, JSON, plain text?

examples/knative/go/kn-go-echo/server.go
codegold79: That's much better. I'll make that change.

examples/knative/go/kn-go-echo/server.go
codegold79: Yes, I like that idea. I'll return the cloud event.

examples/go/kn-echo/README.md
codegold79: I like the idea that @embano1 mentioned with `google/ko`. I'll look into it and see if I can get rid of having to use a Docker image completely.

examples/knative/go/kn-go-echo/server.go
codegold79: Not sure if my answer is what you're looking for. I think you're asking if the `fmt.Print(...)` will print all encoding schemes nicely. Looking at the [String() method on the Event type](https://github.com/cloudevents/sdk-go/blob/a302fc5e5e6a462237f6166905c725921d4da131/v2/event/event.go#L70), it seems like the cloud event SDK library only makes json look nice.

embano1: yeah, you answered my question which was heading in the non-JSON direction, e.g. XML which is currently used in the Tanzu Sources for vSphere.

examples/knative/go/kn-go-echo/server.go
codegold79: @embano1, to what address should I forward the event to? I guess it's the rabbitmq event broker? Would the address be something like `https://rabbit-broker-ingress.default.svc:80`?

codegold79:
> @codegold79 I was wondering if we should use [`ko`](https://github.com/google/ko) to ease the building and deployment of Go related functions? This would remove the need to write and maintain Dockerfiles.

@embano1, @lamw, I've removed `pack` in the build instructions in favor of using `ko`.

examples/knative/go/kn-go-echo/README.md
```diff
@@ -1,83 +1,121 @@
...
-  - `pack` (see: https://buildpacks.io/docs/app-developer-guide/)
-  - Docker
+After installing ko and setting the appropriate environmental variables for it,
+run the following command from the kn-go-echo directory. It will build and push
+the image to your image registry.
 
 ```bash
-IMAGE=vmware/veba-kn-go-echo:latest
-pack build -B gcr.io/buildpacks/builder:v1 ${IMAGE}
+ko publish ../kn-go-echo
```
embano1, line 36: Why the need for `../` syntax?

Can we at least also show `KO_DOCKER_REPO` or use `-L` for local publishing?

examples/knative/go/kn-go-echo/README.md
```diff
@@ -1,83 +1,121 @@
...
+```bash
 $ curl -X POST -i localhost:8080
-HTTP/1.1 204 No Content
-Date: Fri, 02 Apr 2021 21:15:20 GMT
+HTTP/1.1 415 Unsupported Media Type
+Date: Thu, 06 May 2021 21:12:22 GMT
+Content-Length: 0
+```
+
+You can also send a fake cloud event.
```
embano1, line 54: Please only show this example as the other one errors out which could be misleading

examples/knative/go/kn-go-echo/server.go
embano1: You simply return the event like here https://github.com/cloudevents/sdk-go/blob/main/samples/http/responder/main.go#L50

`receiver` is allowed to have different (pre-defined) signatures

examples/knative/go/kn-go-echo/README.md
```diff
@@ -0,0 +1,121 @@
...
+kn trigger create kn-go-echo --broker default --sink ksvc:kn-go-echo
+```
+
+## Build with `ko`
+
+`ko` was created to help Knative developers build images and binaries without
+the need for Dockerfiles nor Makefiles. See more information in [this Knative
+blog](https://knative.dev/blog/2018/12/18/ko-fast-kubernetes-microservice-development-in-go/).
+
+Follow setup instructions at https://github.com/google/ko).
```
embano1, line 29:
```suggestion
Follow setup instructions at https://github.com/google/ko.
```

examples/knative/go/kn-go-echo/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+After installing ko and setting the appropriate environmental variables for it,
```
embano1, line 31:
```suggestion
After installing ko and setting the appropriate environment variables for it,
```

embano1: Can you show which variables? Assuming you're talking about `KO_DOCKER_REPO`?

examples/knative/go/kn-go-echo/README.md
codegold79: I was debating between `.` and `../kn-go-echo`, but I  think I'll go with `.` now.

examples/knative/go/kn-go-echo/README.md
codegold79: Sounds good

examples/knative/go/kn-go-echo/README.md
codegold79: Good catch

examples/knative/go/kn-go-echo/README.md
codegold79: Yup, I was talking about that. I'll include it.

examples/knative/go/kn-go-echo/server.go
codegold79: Nice--thanks for finding that example. I went ahead and implemented responding with the event.

codegold79: I had to add a fix because by returning the same event, it caused an infinite event loop. I changed the event type and added a trigger filter so event types from `kn-go-echo` did not get passed along to `kn-go-echo`.

codegold79:
Thank you @embano1 and @lamw for taking the time to do a thorough review. Because of your feedback, I learned a lot from this simple kn-go-echo function including learning how to use `ko`, using Knative trigger filters, the cloudevents Go SDK, setting up VEBA, and more.

I believe I have implemented all the suggestions, and this PR is again ready for review. Keep those suggestions coming :)

examples/knative/go/kn-go-echo/function.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: Trigger
+metadata:
+  name: kn-go-echo-trigger
+  labels:
+    app: veba-ui
+spec:
+  broker: default
+  filter:
+    attributes:
+      type: com.vmware.event.router/event
```
embano1, line 27: Do we need that filter? (see comment below)

examples/knative/go/kn-go-echo/server.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+}
+
+func receive(ctx context.Context, event cloudevents.Event) *cloudevents.Event {
+	fmt.Printf("***cloud event***\n%s", event)
+
+	response := event
+	// Change the type here and update trigger filter in function.yaml in order
+	// to avoid an infinite event loop.
+	response.SetType("forwarding of com.vmware.event.router/event")
```
embano1, line 31: This is not a well defined type based on the CE spec I would argue. 
For this exercise, you could append `.echo` to the incoming event and ignore any event with that suffix in your logic to avoid loops.

examples/knative/go/kn-go-echo/server.go
```diff
@@ -0,0 +1,33 @@
...
+	log.Println("listening on :8080")
+	if err := client.StartReceiver(ctx, receive); err != nil {
+		log.Fatalf("start receiver: %s", err.Error())
+	}
+}
+
+func receive(ctx context.Context, event cloudevents.Event) *cloudevents.Event {
+	fmt.Printf("***cloud event***\n%s", event)
+
+	response := event
```
embano1, line 28: Please use https://pkg.go.dev/github.com/cloudevents/sdk-go/v2@v2.4.1/event#Event.Clone

examples/knative/go/kn-go-echo/server.go
embano1: So in case of XML the output would not be nice, right?

embano1 (changes requested):

examples/knative/go/kn-go-echo/server.go
codegold79: I haven't tried it, but I'm guessing XML output will be ugly. 

I don't even know if the Cloud SDK http receiver would accept it because it seems I have to use `Content-Type: application/cloudevents+json`. Even `Content-Type: application/json` doesn't seem to work.

examples/knative/go/kn-go-echo/function.yaml
codegold79: I can get rid of the filter in favor of modifying the cloud event in a way that will avoid an infinite event loop.

examples/knative/go/kn-go-echo/server.go
codegold79: Appending `.echo` to the incoming event will make the even no longer a cloud event. What I can do is add a cloud event field called `isEcho` and check that field. If the field is `true`, I will return an empty cloudevent, `event.Event{}`.

If `isEcho == false`, then I'll go ahead and forward the vent.

examples/knative/go/kn-go-echo/server.go
codegold79: I don't even know why I made a copy. It's not like the event being passed in is a pointer. It's just a copy of the original event being sent in. When I send a pointer, it's a pointer to a copy of the event anyway.

examples/knative/go/kn-go-echo/server.go
embano1: I am not following you why appending `.echo` to the event type (string) would break CloudEvents/JSON?

codegold79: > you could append .echo to the incoming event

I think I misunderstood you. I interpreted what you said as, "Append `.echo` to the serialized incoming event". Instead, you meant,

> appending .echo to the event type (string) 

Ok, yes. That is much easier and makes more sense. I can do, 

```go
response.SetType("com.vmware.event.router/event.echo")
```

And instead of using Knative trigger filters to look for this type of event, I can filter using logic in the server.go code.

Is that what you mean?

examples/knative/go/kn-go-echo/server.go
codegold79: I think it's ok to pass along the original event, even if it has some reference types. Let's see what you think with a new commit.

examples/knative/go/kn-go-echo/server.go
embano1: Thx, but please don't make it hardcoded but use append syntax on that string since incoming event type might be different depending on the event.

codegold79: Latest commit is now getting the type with `event.Type()`.

codegold79:
@embano1 Thanks again for all your help. I was able to test the latest changes using vcsim + VMware Event Broker +Kind + Knative. The PR is again ready for your review.

embano1 (approved):
Awesome, thx a ton!

embano1:
Can you please squash the commits before merging @codegold79 ?

codegold79:
> Can you please squash the commits before merging @codegold79 ?

Yup, I can do that. For the future, what do you think about enabling "Squash and Merge" option? Or maybe there was a reason to disable it?

Screenshot from 2021-05-23 10-19-44

embano1:
Squash and merge unfortunately won't give us merge commits as we use when merging :(

codegold79:
> Squash and merge unfortunately won't give us merge commits as we use when merging :(

Gotcha.

I have done a rebase/fixup and combined all changes into one commit. Ready for final review.

embano1 (approved):
Thx so much!

